Issue #29, opened 2019-08-26
Frequency subsampling


ryancoe:
The number of frequency components used in an analysis has a strong effect on the computational load to evaluate a design for (at least) two reasons:

1. NEMOH solver time scales linearly with number of frequencies
2. For solution of the pseudo spectral method (quadratic programming)

Thus, it is worthwhile to minimize the number of frequency components considered. This can be done based on the input spectrum (i.e., the sea state). The basic idea is to:

1. Look for relevant frequency range based on sea state
2. Subsample this range (i.e., select every 5th component)
3. Extend the frequencies up to ~3 harmonics of the highest frequency (this will only matter for extreme loads and things of this sort, not for average power)
4. Use these frequencies when running NEMOH
5. Use these frequencies when solving the PS problem

ssolson:
I have a working version of my implementation of this on my [subSampleW branch](https://github.com/ssolson/WecOptTool/tree/subSampleW).  

To accomplish all the things laid out in this issue it will still need development. I am hoping that @ryancoe & @H0R5E  may be able to pull and run the example.m from my branch to look at what I am currently doing. Lines 36-42 show my implementation. There is a TEMPORARY HARDCODED plot true or false in the removeSpectraTails and downSampleSpectra functions (setting 1 to true at a time shows each function results visually pretty well).

I have made some implementation decisions which adapting from the original is Ryan Made above would look like this:
1. **Load** spectra (example 8 spectra)
2. Pre-Process Data: Iterate over each sea-state. Remove the tails.
3. Pre-Process Data: Iterate over each sea-state. Down-Sample the frequency bins constrained by a maximum error and the minimum number of bins.
        a. Check, via **MAPE** , that the new frequency vector accurately represents the spectra 
        b. **TODO** Check, via moments, that the new frequency vector accurately represents the spectra
3. TODO: make sure to augment for super harmonics

My interest in the discussion would focus on 
1. Do you guys also see this as a pre-processing spectra tool?
2. Should I focus on implementing the current functionality or adding more of the steps laid out here. If add more steps which do I need to focus on?


[test_results.pdf](https://github.com/SNL-WaterPower/WecOptTool/files/4589834/test_results.pdf)


Some other things from Ryan's steps (above) I need to address:
How to check via moments?
How super harmonics?
3. Based on a default frequency spacing, which can be set by the user alternatively, create a finite set of frequencies over the frequency range.
[...]

PR #129, opened 2020-04-22
Consider Sea States Frequency Range in Hydrodynamics

### LINKED ISSUE #77: Hard Coded Frequency
Can someone confirm that this is what we want to be doing here?

https://github.com/SNL-WaterPower/WecOptTool/blob/091be7da6b5cfd07e6876cf6dc29e308dd4da259/toolbox/%2BWecOptLib/%2Bmodels/%2BRM3/%40DeviceModel/getHydrodynamics.m#L85

ssolson:
This PR addresses #77 to consider the device motion in the frequency range of passed sea states (SS). This is achieved by passing SS (struct of sea-states) to getHydrodynamics as a 'varargin'. In the parametric case, SS is passed to a new utility 'seaStatesMinMaxW.m' which returns the min and max frequency in the set of sea states. The frequency vector is then determined by setting a 0.2 bin discretization. This value was chosen because it was the discretization of the current hard-coded frequency. This discretization value will be updated as #29 is addressed. 

 A minor improvement to discuss would be to pass frequency [min, max, nBins] to getNemoh (as opposed to passing a frequency vector ('w') which is currently done) as getNemoh is currently inferring [min, max, nBins] from a passed frequency vector ('w').

The test all pass except runParametric (see attached) because the new solution is outside the tolerance (2.34% > 1% relative tolerance) of the previous solution suggesting the solution is sensitive to this modification. This is important for consideration as #29 is implemented to balance solution accuracy and code speed. Before changing this expected value in the test I wanted to get a review here for discussion. 

Fixes #77 

[test_results.pdf](https://github.com/SNL-WaterPower/WecOptTool/files/4518202/test_results.pdf)


ryancoe:
I had a test fail w/ this (2020a OSX).

[test_results.pdf](https://github.com/SNL-WaterPower/WecOptTool/files/4518816/test_results.pdf)

ssolson:
@ryancoe that is the tolerance test I talk about in the body of my PR. Take a look at the discussion there and let me know what you think.

ryancoe:
Duh... this is acceptable.


I’d like to quantify how much faster this makes the code, will run some tests later

ssolson:
@ryancoe This PR will not speed up the tests (it probably slows it down). This PR considers the passed sea-states to address the previously hardcoded frequency bypassing the min and max frequency over a set 0.2 rad/s discretization

In #29 we discussed checking the RMSE to reduce the number of frequency bins. Once this PR is approved I will start to move on reducing the number of frequency bins used as discussed in #29.

toolbox/+WecOptLib/+utils/seaStatesGlobalW.m
```diff
@@ -0,0 +1,49 @@
...
+% 
+%     WecOptTool is distributed in the hope that it will be useful,
+%     but WITHOUT ANY WARRANTY; without even the implied warranty of
+%     MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+%     GNU General Public License for more details.
+% 
+%     You should have received a copy of the GNU General Public License
+%     along with WecOptTool.  If not, see <https://www.gnu.org/licenses/>.
+
+function w = seaStatesGlobalW(SS, step)
```
H0R5E, line 21: Please write some tests for this function. Particularly, test that the desired step is achieved, and that the bounds of the output are beyond the min and max of SS. Also, please check the defaults, which seems wrong to me.

toolbox/+WecOptLib/+utils/seaStatesGlobalW.m
```diff
@@ -0,0 +1,49 @@
...
+%seaStatesGlobalW Takes a set of sea states and returns a frequency range
+% based on the the global min and  max frequency and a given step size.
+
+gMin=999;
+gMax=-999;
```
H0R5E, line 26: I don't think you need a check on the min and max, you just need to avoid negative frequencies.

H0R5E: And test what happens when negative frequencies are supplied.

ssolson: Trying to understand the bounds on this particular test. It seems to me that checkSpectrum() should be checking for negative frequencies. Should I add a test there on e.g. testValues?

H0R5E: No, please test this function, you never know how it may be used in the future.

H0R5E: It could just chuck an error if a negative frequency is supplied.

ssolson: Because we are passing SS, if negative spectrum should be an error then that error should be caught by checkSpectrum. Why would the tool wait so long to let the user know they have errored? 
I think checkSpectrum should be called at the top of this function to assert this/ throw the error. Do you agree?
I will write another test to look into the other cases.

H0R5E: If you want to modify checkSpectrum and call it, that's fine, and it would be useful to do this to stop the code at an earlier point, I agree. There are a whole bunch of edge cases with this function that you might consider handling and testing (what about a negative step size?) and doing as much as you can now will avoid problems in the future.

H0R5E: I am pushing more on this because this function is in the utils package, which implies wide ranging use, not just in the context of the device model (where we could test the spectrums at an earlier point).

toolbox/+WecOptLib/+utils/checkSpectrum.m
```diff
@@ -85,3 +86,15 @@
...
+function [] = checkPositive(S, idx)
+msg = ['Frequency in Spectrum #%i contains negative values. Frequency'... 
+' values must be positive'];
+ID = 'WecOptTool:invalidSpectrum:notColumnVectors';
+try
+    assert(all(S.w >=0) ,ID,msg, idx)
+catch ME
+    throw(ME)
+end
+end
```
H0R5E, line 100: It's good practice to have a newline at the end of source files.

toolbox/+WecOptLib/+utils/seaStatesGlobalW.m
```diff
@@ -0,0 +1,78 @@
...
+
+function [] = checkStepValue(step)
+msg = 'step must be greater than 0';
+ID = 'WecOptLib:utilityFailure:stepValue';
+try
+    assert(step >0,ID,msg);
+catch ME
+    throw(ME)
+end
+end
```
H0R5E, line 78: It's good practice to have a newline at the end of source files.

toolbox/+WecOptLib/+utils/checkSpectrum.m
ssolson: Will fix. Why is that?

toolbox/+WecOptLib/+utils/checkSpectrum.m
```diff
@@ -49,6 +49,7 @@
     arrayfun(@(Spect,idx) checkFields(Spect, idx), S, inds);
     arrayfun(@(Spect,idx) checkLengths(Spect, idx), S, inds);
     arrayfun(@(Spect,idx) checkCol(Spect, idx), S, inds);
+    arrayfun(@(Spect,idx) checkPositive(Spect, idx), S, inds);
```
H0R5E, line 52: Please add a test for this condition in checkSpectrumTest.m

toolbox/+WecOptLib/+utils/seaStatesGlobalW.m
```diff
@@ -0,0 +1,78 @@
...
+%     along with WecOptTool.  If not, see <https://www.gnu.org/licenses/>.
+
+function w = seaStatesGlobalW(SS, step)
+%seaStatesGlobalW Takes a set of sea states and returns a frequency range
+% based on the the global min and  max frequency and a given step size.
+WecOptLib.utils.checkSpectrum(SS);
+
+checkStepType(step)
+checkStepValue(step)
+
```
H0R5E, line 28: One line of white space is plenty.

toolbox/+WecOptLib/+tests/+unit/checkSpectrumTest.m
```diff
@@ -1,4 +1,5 @@
 
+
```
H0R5E, line 2: What is this for?

toolbox/+WecOptLib/+utils/checkSpectrum.m
```diff
@@ -85,3 +86,15 @@
     throw(ME)
 end
 end
+
+
+function [] = checkPositive(S, idx)
+msg = ['Frequency in Spectrum #%i contains negative values. Frequency'... 
+' values must be positive'];
+ID = 'WecOptTool:invalidSpectrum:notColumnVectors';
```
H0R5E, line 94: This ID is weird

ssolson: good catch

toolbox/+WecOptLib/+utils/checkSpectrum.m
H0R5E: Can cause issues on certain platforms / programs. See https://stackoverflow.com/a/729795/3215152

ssolson:
> Wow, this totally cripples the example, huh? I wonder if it's worth allowing a user chosen discretisation to be set, so that we can run less bins in the example? What do you think? I can add that to this PR if you think it's worth it.

I think that having the option for the user to specify the discretization would be a good option. Currently, the plan on #29 is to reduce the number of frequency bins using an RMSE fit. I think you could either add the user-specified functionality then or now. Let's finish this PR. I'll add a note on #29 to add user-specified step size. Does that work for you?

H0R5E:
How quickly do you think it will take to address #29? I could probably add this in an hour, so it might be worth doing it now, given we have lost all the gains we had made in the testing speed.

ssolson:
2 weeks. In that case, let's add it now.

H0R5E:
OK, will you add that test for checkSpectrum? I will add this user option.

ssolson:
Yes I'm am finishing up the test now.

H0R5E:
OK, so I've added the "freqStep" option to the Parametric geometry, that allows the user to set the discretisation step and I've set it to 0.5 in the example, which has sped the tests up again, a bit. 

I note that RM3DeviceModelParametricCCTest is running much slower as well, but I don't want to change the discretisation on that one as it has a numerical test in it.

Can you see if this works for you?

[test_results.pdf](https://github.com/SNL-WaterPower/WecOptTool/files/4547359/test_results.pdf)

ssolson:
This works great. I think we should be consistent across our naming. So currently you call frequency 'freq' while I use 'w'. Would you prefer me to change the naming to 'freq'?

[test_results.pdf](https://github.com/SNL-WaterPower/WecOptTool/files/4551971/test_results.pdf)

H0R5E:
> I think we should be consistent across our naming. So currently you call frequency 'freq' while I use 'w'. Would you prefer me to change the naming to 'freq'?

That's my bad, I'll fix it.

H0R5E:
I had a look through the code and "freq" is used a lot as a long form of w, so I think it would be a significant job to change them all. I did fix my use of incorrect units though. Are there any specific uses of freq that you think are confusing?

ryancoe:
Please use ‘f’ or ‘freq’ for Hz and ‘w’ for rad/s

ssolson:
Thanks for the nomenclature clarification Ryan. I was thinking we should add something similar to https://mhkit-software.github.io/MHKiT/terminology.html to our documentation. 

Mat I didn't find any use of freq confusing. Was only asking about being consistent.

Are we ready to merge?

H0R5E:
> Please use ‘f’ or ‘freq’ for Hz and ‘w’ for rad/s

Do we ever use the Hz form of frequency? This is the result of a search for "freq" in the source code as it stands:

> example.m	48	parametric = WecOptTool.geom.Parametric(x0, lb, ub, 'freqStep', 0.5);
> complexConjugate.m	21	function [powPerFreq, freqs] = complexConjugate(obj, RM3)
> complexConjugate.m	23	%   Returns power per frequency and frequency bins
> complexConjugate.m	25	% Frequencies
> complexConjugate.m	26	freqs = RM3.w;
> complexConjugate.m	30	powPerFreq = abs(RM3.F0) .^ 2 ./ (8 * real(RM3.Zi));
> damping.m	21	function [powPerFreq, freqs] = damping(obj, RM3)
> damping.m	23	%   Returns power per frequency and frequency bins
> damping.m	30	% Frequencies
> damping.m	31	freqs = RM3.w;
> damping.m	38	% Power per frequency at optimial damping?
> damping.m	39	powPerFreq = 0.5*B_opt *(abs(RM3.F0 ./ (RM3.Zi + B_opt)).^2);
> getHydrodynamics.m	40	%  'freqStep' (float, optional for geomMode='paramtric')
> getHydrodynamics.m	41	%      Sea state frequency discretization (default = 0.2)
> getHydrodynamics.m	55	defaultFreqStep = 0.2;
> getHydrodynamics.m	61	addParameter(p, 'freqStep', defaultFreqStep, validScalarPosNum);
> getHydrodynamics.m	104	freqStep = p.Results.freqStep;
> getHydrodynamics.m	106	% Get global frequency range at freqStep (rad/s) discrtization
> getHydrodynamics.m	107	w = WecOptLib.utils.seaStatesGlobalW(SS, freqStep);
> getMotion.m	30	%           S.w         frequency [rad/s]
> getMotion.m	87	% Wave height in frequency domain
> pseudoSpectral.m	21	function [powPerFreq, freq] = pseudoSpectral(obj, motion)
> pseudoSpectral.m	23	%   Returns power per frequency and frequency bins
> pseudoSpectral.m	38	freq = motion.W;
> pseudoSpectral.m	39	n_freqs = length(freq);
> pseudoSpectral.m	40	powPerFreqMat = zeros(n_ph, n_freqs);
> pseudoSpectral.m	45	[~, phasePowPerFreq] = getPSPhasePower(motion, ph);
> pseudoSpectral.m	47	for ind_freq = 1 : n_freqs
> pseudoSpectral.m	48	powPerFreqMat(ind_ph, ind_freq) = phasePowPerFreq(ind_freq);
> pseudoSpectral.m	53	powPerFreq = mean(powPerFreqMat);
> getPSCoefficients.m	24	% Number of frequency - half the number of fourier coefficients
> getPSCoefficients.m	29	% Frequency vector (re-build)
> getPSPhasePower.m	21	function [pow, powPerFreq] = getPSPhasePower(motion, ph)
> getPSPhasePower.m	24	% power per frequency
> getPSPhasePower.m	80	powPerFreq = Pvec(1:2:end) + Pvec(2:2:end);
> getPSPhasePower.m	81	powPerFreq = powPerFreq * motion.m_scale;
> getPSPhasePower.m	100	assert(WecOptLib.utils.isClose(pow, sum(powPerFreq)))
> DeviceModelTemplate.m	52	%         SS.w         frequency [rad/s]
> DeviceModelTemplate.m	109	% Initialize arrays to hold mu, powSS, powPerFreq, freqs
> DeviceModelTemplate.m	112	powPerFreqs = cell(NSS);
> DeviceModelTemplate.m	113	freqs = cell(NSS);
> DeviceModelTemplate.m	165	[powPerFreq, freq] = obj.complexConjugate(motion);
> DeviceModelTemplate.m	168	[powPerFreq, freq] = obj.damping(motion);
> DeviceModelTemplate.m	171	[powPerFreq, freq] = obj.pseudoSpectral(motion);
> DeviceModelTemplate.m	175	powSS = sum(powPerFreq);
> DeviceModelTemplate.m	180	powPerFreqs{iSS} = powPerFreq;
> DeviceModelTemplate.m	181	freqs{iSS} = freq;
> DeviceModelTemplate.m	190	etc.powPerFreq = powPerFreqs;
> DeviceModelTemplate.m	191	etc.freq = freqs;
> getNemoh.m	23	function [hydro] = getNemoh(r,z,freq,rundir,varargin)
> getNemoh.m	24	% [hydro] = getNemoh(r,z,freq,rundir)
> getNemoh.m	32	%       w       frequency (rad/s) array
> getNemoh.m	36	%       w       frequency array (rad/s)
> getNemoh.m	233	fprintf(fid,'%i	%.2f %.2f		            ! Number of wave frequencies, Min, and Max (rad/s)\n',length(freq),freq(1),freq(end));
> powerPerFreq.m	21	function powerPerFreq(spectra, etc)
> powerPerFreq.m	22	%POWERPERFREQ
> powerPerFreq.m	38	for i = 1 : length(etc.powPerFreq)
> powerPerFreq.m	40	powPerFreq = etc.powPerFreq{i};
> powerPerFreq.m	41	freq = etc.freq{i};
> powerPerFreq.m	42	plot(freq, powPerFreq,'DisplayName',int2str(i))
> powerPerFreq.m	50	powPerFreq = etc.powPerFreq{1};
> powerPerFreq.m	51	freq = etc.freq{1};
> powerPerFreq.m	52	plot(freq, powPerFreq)
> powerPerFreq.m	56	xlabel('Frequency [$\omega$]','Interpreter','latex')
> PlotsTest.m	54	function testPowerPerFreqSingleSpectrum(testCase)
> PlotsTest.m	58	etc.freq = {};
> PlotsTest.m	59	etc.powPerFreq = {};
> PlotsTest.m	60	etc.freq{1} = spectrum.w;
> PlotsTest.m	61	etc.powPerFreq{1} = spectrum.S;
> PlotsTest.m	63	testHandle = @() WecOptLib.plots.powerPerFreq(spectrum, etc);
> PlotsTest.m	68	function testPowerPerFreqMultiSpectra(testCase)
> PlotsTest.m	73	etc.freq = {};
> PlotsTest.m	74	etc.powPerFreq = {};
> PlotsTest.m	78	etc.freq{i} = spectrum.w;
> PlotsTest.m	79	etc.powPerFreq{i} = spectrum.S;
> PlotsTest.m	82	testHandle = @() WecOptLib.plots.powerPerFreq(spectra, etc);
> RM3DeviceModelParametricCCTest.m	35	[S.w, S.S] = WecOptLib.utils.subSampleFreqs(S);
> RM3DeviceModelTest.m	49	[S.w, S.S] = WecOptLib.utils.subSampleFreqs(S);
> RM3DeviceModelTest.m	61	% %[S.w, S.S] = WecOptLib.utils.subSampleFreqs(S);
> RM3DeviceModelTest.m	71	[S.w, S.S] = WecOptLib.utils.subSampleFreqs(S);
> checkSpectrumTest.m	58	eID = 'WecOptTool:invalidSpectrum:negativeFrequencies';
> seaStatesGlobalWTest.m	44	function testCheck_frequecyStepValue(testCase)
> seaStatesGlobalWTest.m	52	function testCheck_frequecyMin(testCase)
> seaStatesGlobalWTest.m	54	shiftInputFrequency = 0.17;
> seaStatesGlobalWTest.m	55	S.w = S.w + shiftInputFrequency;
> seaStatesGlobalWTest.m	76	function testCheck_frequecyMax(testCase)
> seaStatesGlobalWTest.m	78	shiftInputFrequency = 0.7317;
> seaStatesGlobalWTest.m	79	S.w = S.w + shiftInputFrequency;
> seaStatesGlobalWTest.m	100	function testCheck_frequecyRange(testCase)
> seaStatesGlobalWTest.m	102	shiftInputFrequency = 0.7317;
> seaStatesGlobalWTest.m	103	S.w = S.w + shiftInputFrequency;
> seaStatesGlobalWTest.m	128	function testCheck_multiSSFrequecyRange(testCase)
> checkSpectrum.m	30	%       S.w     column vector of frequencies in [rad/s]
> checkSpectrum.m	92	msg = ['Frequency in Spectrum #%i contains negative values. Frequency'...
> checkSpectrum.m	94	ID = 'WecOptTool:invalidSpectrum:negativeFrequencies';
> seaStatesGlobalW.m	22	%seaStatesGlobalW Takes a set of sea states and returns a frequency range
> seaStatesGlobalW.m	23	% based on the the global min and  max frequency and a given step size.
> subSampleFreqs.m	21	function [new_w, new_S] = subSampleFreqs(S, npoints)
> subSampleFreqs.m	22	%subSampleFreq - subsamples sea state and interpolates to three harmonics
> subSampleFreqs.m	29	%        neww = new frequency values
> Normalize.m	8	% data according to ascending frequency.
> Read_NEMOH.m	73	if isempty(strfind(raw{n},'Number of wave frequencies'))==0
> Read_NEMOH.m	75	hydro(F).Nf = tmp{1};  % Number of wave frequencies
> Read_NEMOH.m	76	hydro(F).w = linspace(tmp{2},tmp{3},tmp{1});  % Wave frequencies
> Read_NEMOH.m	343	H(:,j)=sum(local1,2); % H= Kochin function per frequency
> Parametric.m	41	%   freqStep (float):
> Parametric.m	42	%     Angular frequency discretisation step used in NEMOH calculations.
> RM3Study.m	79	%           - S.w: column vector of frequencies in [rad/s]
> plot.m	22	% Plots frequency vs. power for optimized study
> plot.m	45	% Power vs Frequency Plot
> plot.m	46	WecOptLib.plots.powerPerFreq(study.spectra, etc);

It's clear that freq is used to refer to w often. Can I suggest starting another issue to tackle this?

H0R5E:
> Are we ready to merge?

I think we've done enough in this one, and we can fix the nomenclature issue in another one.

ryancoe:
Yes we should address this. Generally the math is easier in rad/s but I often convert to Hz for plotting, as I find it easier to understand.

H0R5E:
> Yes we should address this. Generally the math is easier in rad/s but I often convert to Hz for plotting, as I find it easier to understand.

Opened #132

H0R5E:
@ssolson, do you want me to merge it?

PR #131, opened 2020-04-24
Unique NEMOH File Storage and Cleanup

### LINKED ISSUE #57: Modify NEMOH Data Path
This process is to be developed in the following stages:

1. The path where NEMOH data is stored should be moved local to the source code.

2. The NEMOH data is added to the system temporary folders and retrieved by the user if desired.

3. The NEMOH data is stored in a temporary folder which is deleted on Study object teardown. Similarly the user can retrieve the data prior to deletion if so desired.

H0R5E:
## Description

* This commit enables the RM3Study class to generate unique identifiers for the current MATLAB session and create directory paths for storing NEMOH results based on these identifiers (in the system temporary directory). On class destruction (unless in a parallel worker) any
directory created under that path is removed.

* The user can save the contents of the NEMOH directory associated to the study to a directory using the saveNEMOH method.

* This also allows the reuse of existing NEMOH solutions (mainly for plotting) rather than having to recalculate. To do this the etc struct in getHyrodynamics is serialized into the NEMOH run directories in the study directory. These are then reread after the optimisation to find the directory which matches the optimal solution.

[test_results.pdf](https://github.com/SNL-WaterPower/WecOptTool/files/4529885/test_results.pdf)

Closes #57 

## Checklist:

- [x] Update README section on NEMOH files
- <s>[ ] If `example.m` has been modified, the content / line numbers in  `docs\user\example.rst` are still valid or have been fixed</s>


H0R5E:
@ssolson, the other PR has gone stale (because of making the repo public) so I had to recreate it here. I will answer your comments in that PR then close it. Feel free to pick it back up in this one afterwards.

H0R5E:
Fixed a bug with plotting when using the scalar geometry option. 

[test_results.pdf](https://github.com/SNL-WaterPower/WecOptTool/files/4530163/test_results.pdf)

H0R5E:
Urg, fails for the existing case. Will fix it after dinner...

H0R5E:
Already fixed. Here are the new test results:

[test_results.pdf](https://github.com/SNL-WaterPower/WecOptTool/files/4530373/test_results.pdf)

H0R5E:
Test are faster again now:

[test_results.pdf](https://github.com/SNL-WaterPower/WecOptTool/files/4531069/test_results.pdf)

ssolson:
Picking up on the comments. I think that your stance on comments is a little aggressive. The article you point to makes sense for most code. I will push back against the position for code involving heavy math operations. I think that code performing math should explain the why behind operations the same as any derivation in a journal article. You could say that the math says it all and you would be right, but that doesn't mean its easy to interpret. Giving motivation to an operation is essential in my opinion. Further, I do not believe the why behind performing operations can be encapsulated in the naming of variables and probably should not be, because the variable is not the operation) You could, of course, make a number of intermediate variables based on the name of the operation but I would disagree with this as well. 

Outside of the math, I think I generally agree with this stance on code comments so long as the function/ class containing the code describes well the purpose of the function, what variables it expects, and what it returns using headers (comments).

H0R5E:
> I think that your stance on comments is a little aggressive. The article you point to makes sense for most code. I will push back against the position for code involving heavy math operations. I think that code performing math should explain the why behind operations the same as any derivation in a journal article.

The problem is that comments are not processed, so as code evolves (which it inevitably does), there is no guarantee that the comments will evolve with it (because they don't have to). If you were following [Uncle Bob](https://www.oreilly.com/library/view/clean-code/9780136083238/), then he would pretty much have you writing a function to replace every comment written, so that they are enshrined in the *workings* of the code. Unfortunately, for me, this nuclear option destroys readability, so it's not a goer either. I'm not saying you shouldn't use comments, but if there is a way to make the *code* more descriptive, rather than add a descriptive comment, that is the way to go. 

> Further, I do not believe the why behind performing operations can be encapsulated in the naming of variables and probably should not be, because the variable is not the operation) You could, of course, make a number of intermediate variables based on the name of the operation but I would disagree with this as well.

The "why" of an operation in code is encapsulated in a function name, i.e. `calculateImpedence()`.

toolbox/+WecOptLib/+tests/+data/exampleNEMOH.m
```diff
@@ -0,0 +1,27 @@
...
+%     GNU General Public License for more details.
+% 
+%     You should have received a copy of the GNU General Public License
+%     along with WecOptTool.  If not, see <https://www.gnu.org/licenses/>.
+
+function exampleNEMOH(outputfolder)
+% Unzip example NEMOH files to the given path
+    p = mfilename('fullpath');
+    [pDir, ~, ~] = fileparts(p);
+    zipfilename = fullfile(pDir, "exampleNEMOH.zip");
```
ssolson, line 25: This zip increases the repository size by about 12%. I would like to discuss how this may affect the repository long term if we choose to change this uploaded binary file and then change it. If this is not a concern let me know.

toolbox/+WecOptLib/+tests/+data/exampleNEMOH.m
```diff
@@ -0,0 +1,27 @@
...
+% 
+%     WecOptTool is distributed in the hope that it will be useful,
+%     but WITHOUT ANY WARRANTY; without even the implied warranty of
+%     MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+%     GNU General Public License for more details.
+% 
+%     You should have received a copy of the GNU General Public License
+%     along with WecOptTool.  If not, see <https://www.gnu.org/licenses/>.
+
+function exampleNEMOH(outputfolder)
```
ssolson, line 21: Should we be consistent with the way we capitalize nemoh, Nemoh, and NEMOH? I think we currently lean towards the first 2 depending on if nemoh is the first word in camelcase.

toolbox/+WecOptLib/+tests/+unit/StudyTest.m
```diff
@@ -0,0 +1,96 @@
...
+    % Copy data
+    testDir = fullfile(tempFixture.Folder, "test");
+    testStudy.saveNEMOH(testDir);
+    
+    verifyEqual(testCase, exist(testDir, 'dir'), 7)
+    rmdir(testDir, 's')
+    
+end
+
+function testStudyNoSopyNEMOH(testCase)
```
ssolson, line 80: testStudyNo**C**opyNEMOH?

toolbox/+WecOptLib/+models/+RM3/@DeviceModel/getHydrodynamics.m
```diff
@@ -18,12 +18,13 @@
...
 %     along with WecOptTool.  If not, see <https://www.gnu.org/licenses/>.
 
-function [hydro,rundir] = getHydrodynamics(obj, mode, varargin)
+function hydro = getHydrodynamics(obj, nemohDir, mode, varargin)
 % function [hydro] = getHydrodynamics(mode,...)
 %
 % Returns WEC-Sim hydro structure for RM3.
 %
 % Inputs
+%   nemohDir    Path to folder for NEMOH file storage
```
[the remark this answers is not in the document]
H0R5E: This is set by the RM3Study class, but you would have to set it if you were to use the DeviceModel class directly, yes. I'll probably role this into the geomOptions input from the other PR, once it's merged, as it only relates to the parametric case (for now).

toolbox/+WecOptLib/+models/+RM3/@DeviceModel/getHydrodynamics.m
```diff
@@ -84,7 +84,7 @@
         r2 = varargin{1}(2);
         d1 = varargin{1}(3);
         d2 = varargin{1}(4);
-        [hydro,rundir] = RM3_parametric(r1,r2,d1,d2);
+        hydro = RM3_parametric(r1,r2,d1,d2,nemohDir);
```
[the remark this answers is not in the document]
H0R5E: Yeah, I'll fix it once it's merged.

toolbox/+WecOptLib/+models/+RM3/@DeviceModel/getHydrodynamics.m
```diff
@@ -100,28 +100,10 @@
 
 end
 
-function [hydro,rundir] = RM3_parametric(r1,r2,d1,d2)
+function [hydro] = RM3_parametric(r1,r2,d1,d2,rundir)
```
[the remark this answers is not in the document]
H0R5E: :shrug:

toolbox/+WecOptLib/+tests/+data/exampleNEMOH.m
H0R5E: Given we are storing images for the documentation in this repo, I decided it wasn't too awful to hold this as well. There might be a better way, I'll think about it.

toolbox/+WecOptLib/+tests/+data/exampleNEMOH.m
H0R5E: Hmmm, we probably should be, but I don't know if this is the PR to do it. I strongly believe that NEMOH is how the code is named (even though the github is called Nemoh). I mean, it's in the notice file this way: https://raw.githubusercontent.com/LHEEA/Nemoh/master/NOTICE.txt

ssolson:
>To do this the etc struct in getHyrodynamics is serialized into the NEMOH run directories in the study directory.

okay, now I see the study dir explanation. Sorry.

toolbox/+WecOptLib/+models/DeviceModelTemplate.m
```diff
@@ -184,18 +192,52 @@
...
 
         end
         
     end
     
-end
-
+    methods (Access=protected)
+        
+        function folder = getUniqueFolderPath(obj, path)
```
ssolson, line 217: Do you think that becuase this function is part of the DeviceModel class that `path` should be called studyDir for consistency? If you were to move this function to utils I think a generic 'path' would be appropriate.

toolbox/+WecOptLib/+tests/+data/exampleNEMOH.m
H0R5E: c6cfdda removes the stored solution in favour of reuse of the default example run directories to test the "existing" geometry example.

ssolson:
Hey Mat I am going to pick up my review on this tomorrow. Everything I have tested seems good but working through all the logic is taking some time.

toolbox/+WecOptLib/+models/DeviceModelTemplate.m
H0R5E: No, I think that could add confusion that the getPower argument studyDir is always passed here, which isn't the case - it can take any path. This is effectively a static method, so in theory it could go into the utils module as it stands if it was useful elsewhere.

H0R5E:
Latest test results:

[test_results.pdf](https://github.com/SNL-WaterPower/WecOptTool/files/4553367/test_results.pdf)

toolbox/+WecOptLib/+models/+RM3/@DeviceModel/getHydrodynamics.m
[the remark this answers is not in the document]
H0R5E: With abbe3eb nemohDir is now an optional argument that is only used when in parametric mode. It defaults to ".".

H0R5E:
Tests again:

[test_results.pdf](https://github.com/SNL-WaterPower/WecOptTool/files/4553497/test_results.pdf)

toolbox/+WecOptLib/+models/+RM3/@DeviceModel/getHydrodynamics.m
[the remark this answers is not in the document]
H0R5E: I guess there are less changes using rundir than another name.

toolbox/+WecOptLib/+models/+RM3/@DeviceModel/getHydrodynamics.m
```diff
@@ -37,6 +37,8 @@
 % Name-value Inputs
 %  'spectra' (required for geomMode='paramtric')
```
ssolson, line 38: This is just `mode` now.

toolbox/+WecOptLib/+models/+RM3/@DeviceModel/getHydrodynamics.m
```diff
@@ -37,6 +37,8 @@
...
 %      Sea state description
+%  'nemohDir' (string, optional for geomMode='paramtric')
+%      Directory for NEMOH files (default = ".")
```
ssolson, line 41: is the default here (".")? I thought it was in the temp... folder.  Also, this is just `mode` as well.

toolbox/+WecOptLib/+models/+RM3/@DeviceModel/getHydrodynamics.m
```diff
@@ -37,6 +37,8 @@
...
 %  'freqStep' (float, optional for geomMode='paramtric')
```
ssolson, line 42: `mode='parametric'`

toolbox/+WecOptLib/+models/+RM3/@DeviceModel/getHydrodynamics.m
```diff
@@ -51,13 +53,16 @@
 % (EWTEC2017), Cork, Ireland. 2017.
 %%
 
+defaultNEMOHDir = ".";
```
ssolson, line 56: I see you have added a default NEMOHDir

ssolson: which is in fact here(".")

toolbox/+WecOptLib/+models/DeviceModelTemplate.m
```diff
@@ -22,10 +22,11 @@
     
     methods (Abstract)
         
-        [hydro, rundir] = getHydrodynamics(obj,         ...
-                                           geomMode,    ...
-                                           geomParams,  ...
-                                           varargin)
+        hydro = getHydrodynamics(obj,         ...
+                                 nemohDir,    ...
+                                 geomMode,    ...
```
ssolson, line 27: We need to decide if this is mode or geomMode. E.g. make this input parameter align with the function definition in getHydrodynamics.m

toolbox/+WecOptLib/+models/DeviceModelTemplate.m
ssolson: I disagree that this would cause confusion but clearly there is nothing wrong with the way you have set this up. 

As a static method studyDir **is** always passed here and nothing else. I see that this function could be moved to utils as is; that was my previous point. I believe if you were to move it to utils the input definition should be `path`.

toolbox/+WecOptLib/+tests/+integration/ExampleNEMOHTest.m
```diff
@@ -0,0 +1,106 @@
...
+            
+            cc = WecOptTool.control.ComplexConjugate();
+            study.addControl(cc);
+
+            % Add geometry design variables (existing)
+            d = dir(testCase.Study.studyDir);
+            dfolders = d([d(:).isdir] == 1);
+            dfolders = dfolders(~ismember({dfolders(:).name},   ...
+                                          {'.', '..'}));
+            nemohDir = fullfile(testCase.Study.studyDir, dfolders(1).name);
```
ssolson, line 90: Okay, so the setup class runs the example. Then this test grabs one of the results (e.g. the first index of directories) and then runs the study using an existing geometry.

So what would a failure look like here? Would it be if the existing case failed? I'm trying to understand the scope of that this test catches.

toolbox/+WecOptLib/+utils/isParallel.m
```diff
@@ -0,0 +1,24 @@
...
+%     but WITHOUT ANY WARRANTY; without even the implied warranty of
+%     MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+%     GNU General Public License for more details.
+% 
+%     You should have received a copy of the GNU General Public License
+%     along with WecOptTool.  If not, see <https://www.gnu.org/licenses/>.
+
+function result = isParallel()
+    % Determine if the current process is being run in parallel
+    result = ~isempty(getCurrentTask());
```
ssolson, line 23: getCurrentTask() is a function of PCT. So won't this fail if PCT does not exist?

image

toolbox/+WecOptLib/+tests/+unit/isParallelTest.m
```diff
@@ -0,0 +1,50 @@
...
+end
+
+function testIsParallelTrue(testCase)
+
+    testCase.assumeTrue(WecOptLib.utils.hasParallelToolbox(),   ...
+                        'Parallel toolbox not available')
+    
+    tests = zeros(1, 2);
+    
+    parfor i = 1:2
```
ssolson, line 32: consider i=1:length(tests) incase tests is ever changed

toolbox/+WecOptLib/+tests/+unit/isParallelTest.m
```diff
@@ -0,0 +1,50 @@
...
+
+function testIsParallelFalse(testCase)
+
+    tests = zeros(1, 2);
+    
+    for i = 1:2
+        tests(i) = WecOptLib.utils.isParallel();
+    end
+    
+    verifyEqual(testCase, any(tests), false)
```
ssolson, line 48: Why would you want this to be `any` instead of `all`?

toolbox/+WecOptLib/+models/+RM3/@DeviceModel/getHydrodynamics.m
H0R5E: I think "here" makes the most sense as this method no longer returns the directory where the files are placed, so you couldn't find them again if it placed them in a temporary folder.

toolbox/+WecOptLib/+models/DeviceModelTemplate.m
H0R5E: This definition is just wrong. I will fix it.

toolbox/+WecOptLib/+models/DeviceModelTemplate.m
```diff
@@ -84,11 +88,19 @@
         
         geomOptions = p.Results.geomOptions;
         controlParams = p.Results.controlParams;
+
+        % Create a folder for this call
+        uniqueFolder = obj.getUniqueFolderPath(studyDir);
```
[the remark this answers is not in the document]
H0R5E: No, it's used to store etc as well.

toolbox/+WecOptLib/+tests/+integration/ExampleNEMOHTest.m
H0R5E: Both tests would fail if example.m fails. If example.m runs OK (i.e. generates a completed study object) then the second test is a reproduction of the example using an existing NEMOH solution.

toolbox/+WecOptLib/+utils/isParallel.m
H0R5E: Good point! I'll run it against my serial installation once D3D finishes.

toolbox/+WecOptLib/+tests/+unit/isParallelTest.m
H0R5E: Well, 'all' would return false if only one test was false, whereas I want to check that they are all false, which is only possible with 'any' if they are all false.

toolbox/+WecOptLib/+tests/+integration/ExampleNEMOHTest.m
ssolson: It reproduces but Is it testing that the results are the same?

toolbox/+WecOptLib/+models/+RM3/@DeviceModel/getHydrodynamics.m
H0R5E: Actually it does return the directory in the hydro struct, but I still think it's better for the default to be predictable (and this will only be triggered when using getHydrodynamics directly, which seems unlikely at the moment). 

I guess I could change it to fullfile(tempdir, WecOptTool), but that is perhaps a bit difficult to understand in a docstring. What do you think?

toolbox/+WecOptTool/run.m
```diff
@@ -89,4 +106,30 @@ function run(study, optimOptions)
     study.out.exitflag = exitflag;
     study.out.output = output;
     
+    % Pick up the etc struct for the optimal solution 
+    d = dir(study.studyDir);
+    dfolders = d([d(:).isdir] == 1);
+    dfolders = dfolders(~ismember({dfolders(:).name}, {'.', '..'}));
```
ssolson, line 112: I have seen this code repeated at least 3 times. Should it be a utility?

toolbox/+WecOptLib/+tests/+integration/ExampleNEMOHTest.m
H0R5E: No, that's not the point of this test, it's just to see that the example runs when an existing case is being used.

toolbox/+WecOptTool/run.m
```diff
@@ -89,4 +106,30 @@ function run(study, optimOptions)
...
+        
+        rundir = fullfile(study.studyDir, name{1});
+        etc = load(fullfile(rundir, 'etc.mat'));
+        
+        if isequal(study.out.sol, etc.geomParams)
+            foundMatch = true;
+            break
+        end
+        
+    end
```
ssolson, line 126: This seems like it could be costly for production runs. We may want to investigate the time saved to just run another study on when the number of folders is very large.

H0R5E: Maybe, but this is probably the only way to address #92 if we want to capture the history of more than just the objective function, as in we can load and store all of the etcs.

toolbox/+WecOptTool/run.m
H0R5E: Probably. :cry:

toolbox/+WecOptTool/run.m
ssolson: Okay so the purpose of this is only to grab the results of intermediate optimization results. So we are storing the location of the final run in etc currently and would not need to use this routine for that result?

H0R5E:
Have you finished?

ssolson:
> Have you finished?

I have reviewed everything except the changes to RM3Study.m. I don't think I have the brainpower to review today. Maybe we should have a discussion about the level of the review I should be performing. Am I doing too much? This has been quite a large effort to review all these changes.

toolbox/+WecOptTool/run.m
H0R5E: > So we are storing the location of the final run in etc currently and would not need to use this routine for that result?

No, nothing from the optimisation is kept, so this is needed to recover the etc for the best solution - which is used to make the plots.

I did a quick test of this and it takes 0.007521s to check 5 files, so it would have to check 20000 to be as slow as running the solver again, currently.

toolbox/+WecOptLib/+models/+RM3/@DeviceModel/getHydrodynamics.m
ssolson: My understanding is that `fullfile(tempdir, WecOptTool)` would be the most correct. I agree that it would not be super clear what that meant though. What if the doc string said temporary NEMOH run directory (see etc.hydro.rundir)?

H0R5E:
> Maybe we should have a discussion about the level of the review I should be performing. Am I doing too much? This has been quite a large effort to review all these changes.

I wasn't trying to rush you, I am very grateful for the feeback. It's just that it notifies me on every comment you are making that you've done a review and I'd like to address it all in one, rather than in bits.

ssolson:
Okay, in the future I will do Daily reviews instead of single comments.

toolbox/+WecOptLib/+tests/+integration/ExampleNEMOHTest.m
H0R5E: You can check it with this patch:

```
--- WecOptTool\toolbox\+WecOptTool\run.m
+++ WecOptTool\toolbox\+WecOptTool\run.m
@@ -69,6 +69,7 @@
         study.out.sol = study.geomX0;
         study.out.fval = -1 * negPow;
         study.out.etc = etc;
+        error("Sausages")
         return
         
     end
```

Then testExample will pass but testExistingExample fails.

toolbox/+WecOptLib/+models/DeviceModelTemplate.m
H0R5E: 872faf3 removes the need for this function

toolbox/+WecOptLib/+models/+RM3/@DeviceModel/getHydrodynamics.m
H0R5E: Fixed in 872faf3, to use the built-in `tempname` function as the default

toolbox/+WecOptTool/run.m
H0R5E: Added as a utils function in c55cd98

toolbox/+WecOptLib/+models/+RM3/@DeviceModel/getHydrodynamics.m
```diff
@@ -99,14 +100,20 @@
         
     case 'parametric'
         
+        nemohDir = p.Results.nemohDir;
```
ssolson, line 103: ...NEMOHDir or nemohDir?

toolbox/+WecOptLib/+models/DeviceModelTemplate.m
```diff
@@ -85,8 +85,14 @@
         geomOptions = p.Results.geomOptions;
         controlParams = p.Results.controlParams;
 
-        % Create a folder for this call
-        uniqueFolder = obj.getUniqueFolderPath(studyDir);
+        % Create a folder for this call (and check for race condition)
+        uniqueFolder = tempname(studyDir);
+        [~, ~, message] = mkdir(uniqueFolder);
+
+        if strcmp(message, 'MATLAB:MKDIR:DirectoryExists')
```
ssolson, line 92: what if SUCCESS = 0 and message != 'MATLAB:MKDIR:DirectoryExists'?

H0R5E:
For some reason my No PCT MATLAB died today, and it took me ages to get it running again. Anyway, I managed to run the tests with no PCT and they pass with the one skipped test for isParallel.

[test_results_no_pct.pdf](https://github.com/SNL-WaterPower/WecOptTool/files/4565917/test_results_no_pct.pdf)

Tests are still good with the PCT as well.

[test_results.pdf](https://github.com/SNL-WaterPower/WecOptTool/files/4565921/test_results.pdf)

toolbox/+WecOptLib/+utils/isParallel.m
H0R5E: Tested without PCT for ec1327b

toolbox/+WecOptLib/+models/DeviceModelTemplate.m
```diff
@@ -84,11 +84,25 @@
...
+        % Fix geomOptions if in parametric mode
+        if strcmp(geomMode, 'parametric')
+            geomOptions = [geomOptions, {'nemohDir', uniqueFolder}];
+        end
         
         % WEC-Sim hydro structure for RM3
-        [hydro,rundir] = obj.getHydrodynamics(geomMode,     ...
-                                              geomParams,   ...
-                                              geomOptions{:});
+        hydro = obj.getHydrodynamics(geomMode,      ...
```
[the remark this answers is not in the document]
H0R5E: Here geomMode is the input to getPower and in this context it makes sense to prepend the geom part to separate its meaning from the control options,

toolbox/+WecOptLib/+models/+RM3/@DeviceModel/getHydrodynamics.m
H0R5E: :shrug:

toolbox/+WecOptLib/+models/DeviceModelTemplate.m
H0R5E: Fixing this to check the success status and where I've used it in RM3Study

H0R5E:
Final test results.

With PCT: [test_results.pdf](https://github.com/SNL-WaterPower/WecOptTool/files/4567397/test_results.pdf)

Without PCT: [test_results_no_pct.pdf](https://github.com/SNL-WaterPower/WecOptTool/files/4567398/test_results_no_pct.pdf)

Merging...

PR #135, opened 2020-05-01
make dependencyCheck a function

### LINKED ISSUE #115: Make dependencyCheck.m a function
When `dependencyCheck.m` is run it dumps a load of variables into the workspace:

```
>> who

Your variables are:

allfoundflag                  parallelToolboxFound          
installedNames                parallelToolboxInstalled      
installedProducts             parallelToolboxLicensed       
name                          wafoCheck                     
nemohExistFlag                wafoFunction                  
nemohTestPath                 wafoPath                      
optimizationToolboxInstalled  
optimizationToolboxLicensed  
```

If we make `dependencyCheck.m` into a function, then it will leave the workspace untouched.

ryancoe:
## Description
Previously, with `dependencyCheck` as a script, a bunch of funny variables wound up in the user's workspace. Making `dependencyCheck` a function leaves the workspace tidy and achieves the same result (all relevant info printed to command window).

```
>> clear
>> who
>> dependencyCheck

WecOptTool Dependency Checker
-------------------------------

Required
--------
Optimization Toolbox:          Found
NEMOH:                         Found

Optional
--------
Parallel Toolbox:              Found
WAFO:                          Found

>> who
>> 
```

[test_results.pdf](https://github.com/SNL-WaterPower/WecOptTool/files/4566107/test_results.pdf)

Fixes #115

## Checklist:

### DIFF
--- a/dependencyCheck.m
+++ b/dependencyCheck.m
@@ -18,11 +18,13 @@
 %     You should have received a copy of the GNU General Public License
 %     along with WecOptTool.  If not, see <https://www.gnu.org/licenses/>.
 
-%%%% Purpose: Check that all dependencies are accounted for
-
-%% Init
-
 function dependencyCheck()
+% dependencyCheck()
+%
+% Check that all dependencies are licensed and installed
+%
+% -------------------------------------------------------------------------
+%% Init
 
 allfoundflag = true;
 
--- a/dependencyCheck.m
+++ b/dependencyCheck.m
@@ -19,110 +19,110 @@
 %     along with WecOptTool.  If not, see <https://www.gnu.org/licenses/>.
 
 function dependencyCheck()
-% dependencyCheck()
-%
-% Check that all dependencies are licensed and installed
-%
-% -------------------------------------------------------------------------
-
-allfoundflag = true;
-
-fprintf('\nWecOptTool Dependency Checker\n');
-fprintf('-------------------------------\n');
-
-%% Required Products
-
-fprintf('\n');
-fprintf('Required\n');
-fprintf('--------\n');
-
-%% Optimisation Toolbox
-
-% First Check for ToolBox license
-optimizationToolboxLicensed = license('test', "Optimization_Toolbox");
-
-% Second check if installed
-installedProducts = ver;
-installedNames = {installedProducts(:).Name};
-optimizationToolboxInstalled = false;
-
-for name = installedNames
-    if contains(name, "Optimization Toolbox")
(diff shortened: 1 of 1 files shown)


dependencyCheck.m
```diff
@@ -22,6 +22,8 @@
 
 %% Init
```
H0R5E, line 23: Please remove this comment, as it is now out of context

dependencyCheck.m
```diff
@@ -22,6 +22,8 @@
...
 
+function dependencyCheck()
+
 allfoundflag = true;
 
 fprintf('\nWecOptTool Dependency Checker\n');
```
H0R5E, line 29: Please indent the body of the function

ryancoe: MATLAB's default indentation style (you get this if you select all and press "CONTROL+i") is to not indent functions of this type. I agree that this may not be best for readability, and we can update to `Indent all functions`, but note that there are many files throughout that will need be updated and also that this style preference may be missed by future contributors.

https://www.mathworks.com/help/matlab/matlab_prog/improve-code-readability.html

H0R5E: Yes, I think it's probably worth creating another "policy document" regarding code style. I'm not totally concerned with it being perfect for every, tiny, function (and you would only update a style thing like this when other, more important, changes were being made), but, I do think, for this function indentation is necessary due to its length, to distinguish it from a script. I think the "end" statement, similarly, makes it more clear that this is now a function, even though it's not strictly necessary.

Overall, in a review, the reviewer is always going to bring some of their own preferences to the table, and, as the reviewed, you will decide if you agree or not.

H0R5E:
Also, Mathworks engineers have been fully indenting for a decade (at least)!

https://blogs.mathworks.com/community/2009/05/11/keep-your-code-readable-with-smart-indenting/

dependencyCheck.m
ryancoe: Quite right, I'll make the update and update my MATLAB preferences; if Mathworks is doing it, then it must be good :)

ryancoe:
[test_results.pdf](https://github.com/SNL-WaterPower/WecOptTool/files/4581716/test_results.pdf)

```
>> clear
>> who
>> dependencyCheck

WecOptTool Dependency Checker
-------------------------------

Required
--------
Optimization Toolbox:          Found
NEMOH:                         Found

Optional
--------
Parallel Toolbox:              Found
WAFO:                          Found

>> who
>> 
```

PR #140, opened 2020-05-08
WaveBot

### LINKED ISSUE #70: Arbitrary objective function

### LINKED ISSUE #123: Add Single Body Device Model
As discussed in our last call, it makes a lot of sense to add the ability to model a single body device (probably in heave). This will help us encapsulate code that is shared between all models, and which is shared between similar classes, such as bodies in heave. 

A proposed case study for this model was [WaveBot](https://mhkdr.openei.org/submissions/151), but others candidates may also be viable.

ryancoe:
## Description

Add in `waveBot` class for ongoing paper study. I've used a new programmatic structure here based on our recent discussions @H0R5E and @ssolson where I have a class devoted to the WaveBot. I fully expected to hit a roadblock on this and realize the idea I had was misguided, but I honestly found this to be very straightforward (so far).

This takes over some of the discussion in #125.

Fixes #70 
Fixes #123

## Checklist:

- [ ] All new files contain the GPL header
- [ ] If `example.m` has been modified, the content / line numbers in 
- [ ] add unit tests

## Other To-Do's
- [ ] handle asymmetric constraints fully (in SimResults)


ryancoe:
@H0R5E @ssolson - I don't think this is necessarily ready to merge, but I've appreciate your feedback.

toolbox/+WecOptLib/simRes.m
```diff
@@ -0,0 +1,144 @@
+classdef simRes
```
ssolson, line 1: Spell out results

example_waveBot.m
```diff
@@ -0,0 +1,62 @@
+clc
```
ssolson, line 1: GPL header?

toolbox/+WecOptLib/+models/@waveBot/waveBot.m
```diff
@@ -0,0 +1,350 @@
...
+        function simRes = simPerformance(obj, Spect)
+            % simPerformance    simulates the performance of the design
+            %   finds the maximum power possible for the given design in
+            %   the specified conditions
+            %
+            % Args.
+            %   Spect       wave spectrum (or spectra) defined in the style
+            %               of WAFO
+            % Returns
+            %   simRes      structure (TODO - make this a class?)
```
ssolson, line 206: I think you did this TODO? remove the comment either way and start a discussion if you want to discuss.

waveBot_caseA.m
```diff
@@ -0,0 +1,48 @@
+clc
```
ssolson, line 1: Is this file to be published in the repository or just for the paper? GPL header?

toolbox/+WecOptLib/simRes.m
ryancoe: what do you mean?

toolbox/+WecOptLib/+models/@waveBot/waveBot.m
ryancoe: completed w/ fd7787ca91011cfdf56938d09c8457c5861811e6

waveBot_caseA.m
ryancoe: not sure... open to discussion

toolbox/+WecOptLib/simRes.m
ssolson: this means simResults right? Or is it simResonance? or simResource? I think the clarity of results is worth the extra 3 characters

toolbox/+WecOptLib/+models/@waveBot/waveBot.m
```diff
@@ -0,0 +1,350 @@
...
+                    % position
+                    pos = u ./ (1i * obj.hydro.w);
+                    
+                    % PTO force
+                    Fpto = -Zpto .* u;
+                    
+                    % power
+                    pow = 0.5 * Fpto .* conj(u);
+                    
+                case 'PS' % -----------------------------------------------
```
ryancoe, line 295: @gbacelli - Can you please help by implementing the single body PS here? You can reference/copy-paste from the following for the RM3:

 - [`getPSCoefficients.m`](https://github.com/SNL-WaterPower/WecOptTool/blob/master/toolbox/%2BWecOptLib/%2Bmodels/%2BRM3/getPSCoefficients.m) - this function creates the variables that are used
 - [`getPSPhasePower.m`](https://github.com/SNL-WaterPower/WecOptTool/blob/master/toolbox/%2BWecOptLib/%2Bmodels/%2BRM3/getPSPhasePower.m) - this function forms the objective function for optimization. 
 - [`pseudoSpectral.m`](https://github.com/SNL-WaterPower/WecOptTool/blob/master/toolbox/%2BWecOptLib/%2Bmodels/%2BRM3/%40DeviceModel/pseudoSpectral.m) - this function solves the PS optimization problem (and does so `n` times to average for phasing)

If done correctly, I believe [`getPSPhasePower.m`](https://github.com/SNL-WaterPower/WecOptTool/blob/master/toolbox/%2BWecOptLib/%2Bmodels/%2BRM3/getPSPhasePower.m) and [`pseudoSpectral.m`](https://github.com/SNL-WaterPower/WecOptTool/blob/master/toolbox/%2BWecOptLib/%2Bmodels/%2BRM3/%40DeviceModel/pseudoSpectral.m) can be the same for all devices.

toolbox/+WecOptLib/simRes.m
ryancoe: renamed to `simResults` w/ 2317e2bc3569d8f97c98ea01b22f3f020c98cda7

ssolson:
Thoughts on the example_waveBot results:
I like the table. 
I think F-count is the total number of objective function calls?
What is Feasibility and why is it 0?
Is the 1st order optimality compared to the MC runs? If so I doubt the user will always have an MC to compare against.
Not sure what Norm of the step is.

What is "Current Point" (e.g. consider renaming)? Why is it a bar graph? I think this is updated each objective function iteration e.g. each F-count. If so could you do an F-count vs Result and a scatter plot?
I like seeing the Function value as a function of iteration.
I like seeing the optimization function vs the MC but again do not thing this will always be an option for the user.  Of course, I see this is an ad-hoc plot and not built into the results

image

ryancoe:
@ssolson - On your comments about the `example_waveBot` results - most of this is built into MATLAB (the command line output, plot functions, etc.). Any questions you have about the meanings are probably best found on the MATLAB documentation. For tweaks, we can of course come up with some revised versions (I agree with you about plotting the history of the solution).

H0R5E:
> Also in a discussion with Mat today on D3D it came up that we should probably tag a version before we start implementing any major changes especially as a reference point before the MEC presentation.

Yeah, if we are going to present the v0.1 concept then I think it would be a good idea to do another release before we remove the existing WecOptTool UI. My expectation is that whatever the outcome of this is, it will overwrite WecOptTool to become the public API.

ryancoe:
> My suggestion is to try and adapt the structure of DeviceModelTemplate to look more like the waveBot class and then port the waveBot code into a concrete DeviceModelTemplate class next to the existing RM3 model (and fix that model as required for the modifications).

Yes, I think this is the idea... for me, making the one-off `WaveBot` class was helpful for me to work through the structure and steps. At the same time, I really do like the workflow for performing a study demonstrated in [`waveBot_caseA`](https://github.com/SNL-WaterPower/WecOptTool/blob/2317e2bc3569d8f97c98ea01b22f3f020c98cda7/waveBot_caseA.m).

A lot of the process for me was simply understanding what had already been done for v0.1 a bit more completely. For instance, I think that the following methods in `WaveBot` class (not exhaustive) can be common between devices (i.e., part of a super class `DeviceModelTemplate`):

1. Some of the components of the `init` method
https://github.com/SNL-WaterPower/WecOptTool/blob/2317e2bc3569d8f97c98ea01b22f3f020c98cda7/toolbox/%2BWecOptLib/%2Bmodels/%40waveBot/waveBot.m#L50

2. The post-processing section of `runHydro`
https://github.com/SNL-WaterPower/WecOptTool/blob/2317e2bc3569d8f97c98ea01b22f3f020c98cda7/toolbox/%2BWecOptLib/%2Bmodels/%40waveBot/waveBot.m#L80

3. The `CC` and `P` cases from `simPerformance`. Also some parts of the `PS` steps (execution of solver).
https://github.com/SNL-WaterPower/WecOptTool/blob/2317e2bc3569d8f97c98ea01b22f3f020c98cda7/toolbox/%2BWecOptLib/%2Bmodels/%40waveBot/waveBot.m#L197

4. All of the protected methods
https://github.com/SNL-WaterPower/WecOptTool/blob/2317e2bc3569d8f97c98ea01b22f3f020c98cda7/toolbox/%2BWecOptLib/%2Bmodels/%40waveBot/waveBot.m#L318

As I'm sure you're noting, I have really been reinventing the wheel with this `WaveBot` class, as the reusable components I've identified here closely follow what's been done with `DeviceModelTemplate`. I think that the structure would look something like this (ignoring the `simResults`, etc. for expediency):

image

What does this mean for someone wanting to make a new device:

 - if the user's device looks kinematically just like an existing device, you'd need to copy-paste (or just edit in place) and change the code for creating the geometry. I suppose this means that the classes `RM3` and `WaveBot` could eventually become archetypes (`twoBodyPa` and `oneBodyPa`, respectively).
 - if the user's device does not fit kinematically into one of the existing devices, they'd have to create a class with the proper interfaces (eventually, we aim to have some semi-automated process for this based on the discussion in #66, but this may be a way off still)

toolbox/+WecOptLib/+models/@waveBot/waveBot.m
```diff
@@ -0,0 +1,350 @@
+classdef waveBot < handle
```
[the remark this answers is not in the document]
ryancoe: fixed w/ 5acef655fd47d9f001de0b452c8f32947b741326

toolbox/+WecOptLib/+models/@waveBot/waveBot.m
```diff
@@ -0,0 +1,350 @@
...
+    %     the License, or (at your option) any later version.
+    %
+    %     WecOptTool is distributed in the hope that it will be useful, but
+    %     WITHOUT ANY WARRANTY; without even the implied warranty of
+    %     MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+    %     General Public License for more details.
+    %
+    %     You should have received a copy of the GNU General Public License
+    %     along with WecOptTool.  If not, see
+    %     <https://www.gnu.org/licenses/>.
```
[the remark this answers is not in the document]
ryancoe: I did have this at the top at first, however, if you have the GPL stuff at the top, you don't get the right result in MATLAB:

```
>> help WecOptLib.models.WaveBot
  Copyright 2020 National Technology & Engineering Solutions of Sandia,
  LLC (NTESS). Under the terms of Contract DE-NA0003525 with NTESS, the
  U.S. Government retains certain rights in this software.
 
  This file is part of WecOptTool.
 
      WecOptTool is free software: you can redistribute it and/or
      modify it under the terms of the GNU General Public License as
      published by the Free Software Foundation, either version 3 of
      the License, or (at your option) any later version.
 
      WecOptTool is distributed in the hope that it will be useful, but
      WITHOUT ANY WARRANTY; without even the implied warranty of
      MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
      General Public License for more details.
 
      You should have received a copy of the GNU General Public License
      along with WecOptTool.  If not, see
      <https://www.gnu.org/licenses/>.

    Documentation for WecOptLib.models.WaveBot
```

By putting the GPL stuff below the documentation, you get this (this is especially helpful with OOP, because you can call help on an instance of a class):

```
>> help WecOptLib.models.WaveBot
  WaveBot   WECs based on the Sandia "WaveBot" device
  The WaveBot is a model-scale wave energy converter (WEC) tested in
  the Navy's Manuevering and Sea Keeping (MASK) basin. Reports and
  papers about the WaveBot are available at advweccntrls.sandia.gov.
 
  waveBot Properties:
    controlType - determines type of controller ('P', 'CC', or 'PS')
    geomType - determines way device geometry is set by user
    w - determines frequencies for evaluation [rad/s]
    studyDir - location to run BEM
    hydro - structure created by Read_NEMOH
 
  waveBot Methods:
    runHydro - defines geometry and runs BEM
    plot - plots geometry cross section
    simPerformance - simulates the performance of the design
 
  See also WecOptLib.nemoh.getNemoh, Read_NEMOH
 
  Copyright 2020 National Technology & Engineering Solutions of Sandia,
  LLC (NTESS). Under the terms of Contract DE-NA0003525 with NTESS, the
  U.S. Government retains certain rights in this software.
 
  This file is part of WecOptTool.
 
      WecOptTool is free software: you can redistribute it and/or
      modify it under the terms of the GNU General Public License as
      published by the Free Software Foundation, either version 3 of
      the License, or (at your option) any later version.
 
      WecOptTool is distributed in the hope that it will be useful, but
      WITHOUT ANY WARRANTY; without even the implied warranty of
      MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
      General Public License for more details.
 
      You should have received a copy of the GNU General Public License
      along with WecOptTool.  If not, see
      <https://www.gnu.org/licenses/>.

    Documentation for WecOptLib.models.WaveBot
```

Looking around more... `WAFO` also uses GPL, but does not put the GPL listing in each source file (same for [`MSS`](https://github.com/cybergalactic/MSS/blob/master/FDI/EditAB.m)). `WEC-Sim` (Apache) doesn't put the listing in each file either.

toolbox/+WecOptLib/+models/@waveBot/waveBot.m
```diff
@@ -0,0 +1,350 @@
...
+                w = 2*pi*linspace(0.05, 2, 50)';
+            end
+            
+            obj.controlType = controlType;
+            obj.geomType = geomType;
+            obj.w = w;
+            obj.makeStudyDir();
+        end
+        
+        function runHydro(obj, geomDef)
```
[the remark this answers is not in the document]
ryancoe: To make sure I understand correctly, you're saying that calling `runHydo` separately from `simPerformance` is problematic?

If I'm understanding that correctly, my response is that `runHydro` is indecent of the wave spectrum (ignoring #29 for now) - effectively the hydrodynamics of your device depend only on the geometry (again ignoring local linearization and the stuff from #29).

This is illustrated in `WaveBot_caseA`. I could actually have created a single device, called `runHydro`, and then duplicated it to have different controllers, e.g., (since in this case Nemoh is the main computational burden, this will run twice as fast)

```
>> a = WecOptLib.models.WaveBot('CC',geomMode,w);
>> a.runHydro(1)
>> b = a;
>> b.controlType = 'P';
```

I suppose another way to handle this is the `existing` flag for the geometry... I guess I like having both options available.

> If it's useful for the user to call this as a separate stage, the create some output that is required by simPerformance to make it explicit.

Not sure how to do this or exactly what you mean... I'm currently throwing an error if `obj.hydro` is empty when `simPerformance` is called:

https://github.com/SNL-WaterPower/WecOptTool/blob/2317e2bc3569d8f97c98ea01b22f3f020c98cda7/toolbox/%2BWecOptLib/%2Bmodels/%40waveBot/waveBot.m#L222

toolbox/+WecOptLib/+models/@waveBot/waveBot.m
```diff
@@ -0,0 +1,350 @@
...
+            plot(ax,obj.geom.r, obj.geom.z,'o-')
+            
+            legend('location','southeast')
+            xlabel('r [m]')
+            ylabel('z [m]')
+            ylim([-Inf, 0])
+        end
+        
+        
+        function simResults = simPerformance(obj, Spect)
```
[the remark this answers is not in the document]
ryancoe: > I don't see why the spectrum is required at this stage and not as an input to the class.

I see the device as a fixed object (like in real life) that can be **_fed_** any number of spectra. This may actually be the single largest difference between what I've done here vs. the structure in v0.1. In v0.1, we would form a `study`. I found that restrictive and that was the main motivation for this PR. 

> Are we expecting to loop over the spectra in the optimiser?

yes, I have been lazy and not implemented the capability to handle multiple spectra in `simPerformance`, but this is relatively easy

toolbox/+WecOptLib/+models/@waveBot/waveBot.m
```diff
@@ -0,0 +1,350 @@
...
+            % simPerformance    simulates the performance of the design
+            %   finds the maximum power possible for the given design in
+            %   the specified conditions
+            %
+            % Args.
+            %   Spect       wave spectrum (or spectra) defined in the style
+            %               of WAFO
+            % Returns
+            %   simResults  object containing simulation results with
```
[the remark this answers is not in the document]
ryancoe: 28b74741a17731bc28151fc93b40feeaafad6cb6

toolbox/+WecOptLib/+models/@waveBot/waveBot.m
```diff
@@ -0,0 +1,350 @@
...
+            %               w       frequency vector
+            %               eta     complex wave elevation spectrum
+            %               Fe      complex excitation spectrum
+            %               pow     real power spectrum
+            %                       (tot_avg_pow = sum(simResults.pow)
+            %               u       complex velocity spectrum
+            %               Zpto    PTO impedance
+            %               Fpto    complex PTO force spectrum
+            %
+            % See also JONSWAP, WecOptLib.simResults
```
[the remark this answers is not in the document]
ryancoe: because it produces the structure `spect` that will be used... yeah, you're right that it wasn't very clear - updated docs w/ b79cc4b98879e49e1dba29ddefd45bb04eddb607

toolbox/+WecOptLib/simResults.m
```diff
@@ -0,0 +1,136 @@
+classdef simResults
```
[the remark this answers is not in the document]
ryancoe: c9863515b1644e9d207583ada56f52e90de72ab9

toolbox/+WecOptLib/simResults.m
```diff
@@ -0,0 +1,136 @@
...
+        Fpto
+        pow
+        name
+    end
+    methods
+        function obj = simResults(name)
+            obj.name = name;
+        end
+        
+        function plotTime(obj ,t)
```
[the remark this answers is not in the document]
ryancoe: yes, I knew this was dirty when I did it but was just wanting to get something working.

I guess I justified this implementation by the fact that instances of `SimResults` should only be created by internal methods (e.g., `simPerformance`), not by the user... agree that we should do something more robust

H0R5E:
OK, so I have a very short time to write this down, and perhaps it warrants another chat, but my main thought is that, if we are using objects to define real concepts, an actual device doesn't come into existence until it has a size defined. This is causing the temporal coupling, because WecDevice morphs into something else once you call runHydro. In reality, we start with a device _blueprint_ and then we make a device from it by defining a size. So, I think we need two classes, one for the blueprint and one for the created device like so:

WecOptTool

I've added some of the encapsulation ideas we talked about in here as well, but, basically, the user would be expected to build the blueprint and then the rest would be done by the toolbox. What do you think?

ryancoe:
@H0R5E - I think you may be on to something... so programmatically it would be something like this?

```
myBp = WecOptLib.models.RM3(<controlType>,<geomType>, ...);
myWec = myBp.makeDevice(<size or geometry params>);
simResults = myWec.simPerformance(S);
```

H0R5E:
Almost, I think it would make sense to define what type of geometry and control would be called in different places, like so:

```
myBp = WecOptLib.models.RM3();
myWec = myBp.makeDevice(<geomType>, <size or geometry params>);
simResults = myWec.simPerformance(S, <controlType>);
```

This would give you more flexibility for reuse of your objects and would probably make it easier to define the encapsulated functions, so for makeMesh, it might have a standard definition, like

```
mesh = makeMesh(<geomType>, <size or geometry params>)
```

H0R5E:
Alternatively you could define a control type when you make the device, which could be changed later:

```
myBp = WecOptLib.models.RM3();
myWec = myBp.makeDevice(<geomType>, <size or geometry params>,  <controlType>);
simResults = myWec.simPerformance(S);
myWec.setControl(<anotherControlType>)
simResults2 = myWec.simPerformance(S);
```

toolbox/+WecOptLib/+models/@waveBot/waveBot.m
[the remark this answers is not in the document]
H0R5E: As discussed in our call, not using the GPL notice in a file is probably incorrect application of the license. To quote [How to use GNU licenses for your own software](https://www.gnu.org/licenses/gpl-howto.en.html)

> - Put a license notice in each file.

I had a look at how MATLAB does this and the rules are as follows:

* If there is no docstring, the copyright notice is placed at the bottom of the file
* If a docstring exists, the copyright notice is placed after the docstring but, crucially, there is a blank line between the docstring and the copyright notice. E.g.:

    ```
    classdef ComplexConjugate < WecOptTool.control.AbsControl
        % Provides complex congugate control
        %
        % Seeks optimal power absorption, see `Falnes2002`_ for technical 
        % details.
        % 
        % .. _Falnes2002:
        %     https://www.cambridge.org/core/books/ocean-waves-and-\
        %     oscillating-systems/8A3366809DE5C1F916FF87F36C55C459
        
        % Copyright 2020 National Technology & Engineering Solutions of Sandia, 
        % LLC (NTESS). Under the terms of Contract DE-NA0003525 with NTESS, the 
        % U.S. Government retains certain rights in this software.
        %
        % This file is part of WecOptTool.
        % 
        %     WecOptTool is free software: you can redistribute it and/or 
        %     modify it under the terms of the GNU General Public License as 
        %     published by the Free Software Foundation, either version 3 of 
        %     the License, or (at your option) any later version.
        % 
        %     WecOptTool is distributed in the hope that it will be useful,
        %     but WITHOUT ANY WARRANTY; without even the implied warranty of
        %     MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
        %     GNU General Public License for more details.
        % 
        %     You should have received a copy of the GNU General Public License
        %     along with WecOptTool.  If not, see 
        %     <https://www.gnu.org/licenses/>.
    ```

The blank line stops the help system picking up the copyright declaration and sphinx doesn't read it either. I suggest we follow a similar approach.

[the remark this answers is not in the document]
ryancoe: I will address the files specific to this PR, but started a dedicated issue (#142) to handle throughout source.

ryancoe:
I think it's time to go ahead and try one of these approaches. It won't be too hard to switch if we find something we don't like.

H0R5E:
> I think it's time to go ahead and try one of these approaches. It won't be too hard to switch if we find something we don't like.

I'm happy to put something together, but it will take me a couple of days, if that's OK?

toolbox/+WecOptLib/+models/@waveBot/waveBot.m
ryancoe: @gbacelli 👉

toolbox/+WecOptLib/+models/@WaveBot/WaveBot.m
```diff
@@ -0,0 +1,557 @@
...
+            
+            % y is a column vector containing [vel; u] of the
+            % pseudospectral coefficients
+            tmp = reshape(y,[],2);
+            x1hat = tmp(:,1);
+            uhat = tmp(:,2);
+            Phat = 1/2 * x1hat .* uhat;
+            
+%             % find the spectra
+            ps2spec = @(x) (x(1:2:end) +  x(2:2:end));  % TODO - probably make this a global function
```
ryancoe, line 493: @gbacelli - I think I should be making these frequency vectors complex at this stage, but the obvious solution (xInFreq = `complex(xInPs(1:2:end) +  xInPs(2:2:end))`) doesn't seem to work...

ryancoe:
@gbacelli - I fixed my initial issue with the power spectrum (Parseval...) with ffeb742342817737c9d97e1e07a0ca46cdf6177f, but I'm now noticing that the limits, e.g., on force, don't seem to be working symmetrically... Here's a minimum working example below. As you can see, the positive force is limited to 50N as desired, but not the negative...

```Matlab
clc
clear
close all

geomMode = 'scalar';
lambda = 1;
w = 2*pi*linspace(0.05, 2, 50)';
a = WecOptLib.models.WaveBot('PS',geomMode,w);
a.runHydro(lambda);
a.delta_Fmax = 50;
a.delta_Zmax = 1e10;

% regular wave
Hm0 = 0.125;
Tp = 1/0.625;
gamma = 3.3;
S = jonswap(a(1).w,[Hm0, Tp, gamma],0);
[~,idx] = min(abs(S.w - 2*pi/Tp));
Sn = S;
Sn.S = Sn.S * 0;
Sn.S(idx) = S.S(idx);
S = Sn;
clear Sn

%%

rng(3)
r = a.simPerformance(S);  

summary(r)

r.plotFreq()
r.plotTime(0:0.01:10)
```
image

image


Here's the line where that inequality matrix is created: https://github.com/SNL-WaterPower/WecOptTool/blob/b20bfb660ca32dfe936774565ac05f1eeda10cfe/toolbox/%2BWecOptLib/%2Bmodels/%40WaveBot/WaveBot.m#L433

WaveBot_caseA.m
```diff
@@ -0,0 +1,101 @@
...
+% of the basic differences between these three controller types:
+%
+% CC    complex conjugate control
+% P     proportional damping
+% PS    pseudo-spectral numerical optimal control
+%
+% This case study is detailed in the following paper:
+% 
+% @Article{WecDesignOptimizationTool,
+%   author       = {Ryan G. Coe and Giorgio Bacelli and Sterling Olson and
```
ssolson, line 15: Mathew (one t)

example_waveBot.m
```diff
@@ -0,0 +1,62 @@
...
+lambdas = 0.25:0.25:2;
+mcres = arrayfun(@(x) myWaveBotObjFun(x,a,S), lambdas);
+
+%% via a solver
+
+x0 = 1;
+A = [];
+B = [];
+Aeq = [];
+Beq = [];
```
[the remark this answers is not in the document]
ryancoe: No plans to revise, in fact plan to specifically not revise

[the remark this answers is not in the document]
ssolson: Please ignore the previous comment.

H0R5E:
Isn't this superseded by #159? Is there any point in @ssolson going over this?

ssolson:
> Isn't this superseded by #159? Is there any point in @ssolson going over this?

Mat the issue existed there as well. Ryan set me straight on the correct repository to look at yesterday though and I am not reviewing this. 

I think we should close this PR though

ryancoe:
now under #161

Issue #126, opened 2020-04-22
Modularisation and Data Transfer


H0R5E:
This is a proposal for an architecture for the library code (i.e WecOptLib) that will provide three benefits. These are:

1. Separation (modularisation) of the major functional components of the code, such that they become solely data coupled. This then allows work to be done within each components without the need to modify or having any effect upon other components.
2. Reliable, documented data structures (data transfer objects) are used to transfer data between the functional components. This provides certainty for a consumer of this data, be that the user or another functional component.
3. Device independence. This framework should not be abstracted per device, provided the devices given are within scope. Abstraction of the functional components for device classes may still be beneficial.

A schematic of this architecture is:

![WecOptTool Components](https://user-images.githubusercontent.com/5088535/79973129-ea56c900-848e-11ea-8b18-e55241bcc547.png)

Data Transfer Objects
-----------------------

At the top of the diagram are the data transfer objects (DTOs). These are fixed APIs that carry data between the major functional components of the tool. An implementation of these would be a public facing class that contains a Struct and a method for populating the struct to check the validity of the inputs.

The three DTOs defined here are:

1. *Environment / Device*: This is populated by the user and the optimiser to define the environmental and device parameters. All inputs required to run the "behaviour" component must be supplied here.
2. *Loadings*: This is populated by the "behaviour" component and contains all the data that is required for the "motion control" component to calculate and control the device motions.
3. *Motions*: This is populated by the "motion control" components and contains all the data required to build a cost function in the "evaluation component".

Functional Components
--------------------------

The lower half of the diagram contains the key functional components of the code and defines the points of separation provided by the DTOs. Ultimately, these components take a DTO as input and return a DTO as output - how they are implemented within the block is not important. However, it is probably useful to structure these in such a way that they are abstracted for different cases, i.e. they are structured to avoid code repetition and lots of if statements. 

The three components defined are:

1. *Behaviour*: this component is responsible for calculating the loadings on the device by calling an external tool (e.g. NEMOH). It can be abstracted for different external tools and may also be useful to abstract it for different device classes (like single body vs multi-body, fixed or floating, etc).
2. *MotionControl*: this component calculates the controlled motion of the device. It is likely to be abstracted per controller and/or per device type, if necessary.
3. *Evaluation*: this component is used for defining a cost function and is likely to be user facing. It is a thin wrapper to allow access to the DTOs so that a cost function can be derived. Some prebaked examples could also be provided.

Device Model
---------------

[...]

PR #143, opened 2020-05-19
Sub sample w

### LINKED ISSUE #29: Frequency subsampling

ssolson:
## Description

This draft PR address the discussion of #29 to set up preprocessing tools to allow the user to preprocess a spectra prior to optimizing. This is done through the introduction of 3 utilities:

1. removeSpectaTails.m
2. downSampleSpectra.m
3. resampleSpectra.m

The use of the new tools is shown in example.m. 

Two new plotting functions have been added to allow users to easily visualize changes made to a spectra from using the 3 functions above. Namely:
1. compareNoTailsSS.m
2. compareSpectra.m

In order to perform error checking across functions an `errorCheck` folder has been added which provides an easy way to check inputs etc. using e.g. `assertGreaterThan` or `assertPositiveInteger`.

Currently, I am writing tests for the new functions but I wanted to get this draft PR started as this is becoming a bit of an omnibus.

Closes #29

## Checklist:


example.m
```diff
@@ -28,13 +28,46 @@
...
+% addParameter(p,'wMin',0.11)
+% addParameter(p,'wMax',4.1)
+
+% Automatically downsample the set dw based on down-sampled spectrum
+maxError=1;
+downSampledS = WecOptLib.utils.downSampleSpectra(noTailsS, maxError, minBins);
+WecOptLib.plots.compareSpectra(noTailsS, downSampledS, 'downSampleError');
+downSampledS = WecOptLib.utils.getMeanDw(downSampledS);
+% Create spectrum using S.dw and S.w
+resampledS = WecOptLib.utils.resampleSpectra(downSampledS);
```
ryancoe, line 66: Add optional argument for number of super harmonics with default of 6.

example.m
```diff
@@ -28,13 +28,46 @@
...
+%S = WecOptLib.tests.data.exampleSpectrum();
 
 %% Or load an example with multiple sea-states (8 differing spectra)
 S = WecOptLib.tests.data.example8Spectra();
 
+%% Pre-process spectra to optimize the number of frequency bins
+
+% Find meaningful part of the Spectra by removing tails of the spectra
+% Set a percentage tolerence 
+tailTolerence = 1;
```
ryancoe, line 40: I prefer to write 1% as 0.01

example.m
```diff
@@ -28,13 +28,46 @@
...
+% Specify a minimum number of frequency Bins
+minBins = 10;
+% Remove the tails
+noTailsS = WecOptLib.utils.removeSpectraTails(S, tailTolerence, minBins);
+% Plot a comparison of the original and resultant spectra
+%WecOptLib.plots.compareNoTailsSS(S, noTailsS)
+
+% Resample the spectra based on frequency step size
+% Set a frequency step size manually 
+dw=0.33;
```
ryancoe, line 50: Pass `dw` as an argument

ssolson:
> use MATLAB Argument Validation framework where possible:
> 
> https://www.mathworks.com/help/matlab/matlab_prog/function-argument-validation-1.html

@ryancoe looking into this more this is not what we want really. First of all, MATLAB will first attempt a conversion for set data size, type, and class. Secondly, the optional assert statements must be written by you as a function at the bottom of the code which then recreates the problem I am solving by making the error checking folder. However, it seems that this functionality can be used to set default values which is very useful. I currently plan to finish out the errorCheck folder based on my review of this MATLAB functionality. Let me know what you think.

H0R5E:
> use MATLAB Argument Validation framework where possible:
> 
> https://www.mathworks.com/help/matlab/matlab_prog/function-argument-validation-1.html

I didn't even know this existed until today. This is when MATLAB starts to look more like Java with a very typed language kind of feature, particularly with the casting it tries to do. I should imagine this can be super useful (I think I will try it myself), but should be used with care.

ssolson:
I was able to implement and use the MATLAB function validation using this table:

https://www.mathworks.com/help/matlab/matlab_oop/property-validator-functions.html

ssolson:
Example of the new plotting function compareNoTailsSS as requested by @ryancoe.

Side note is that my legend is not showing up properly. I see the same behavior for the `powerPerFreq()` plot. Anyone else having this issue?

image

ssolson:
Same as above for `compareSpectra`

image

ssolson:
@H0R5E my tests are failing on two NEMOH tests (see attached). Is this the same error Kelly said the Jenkins server was returning?
[test_results.pdf](https://github.com/SNL-WaterPower/WecOptTool/files/4726402/test_results.pdf)

H0R5E:
No, just looks like a bug in your `example.m` to me:

```
Unrecognized function or variable 'downSampledS'.
Error in example (line 52)
resampledS = WecOptLib.utils.resampleSpectra(downSampledS, dw); %Not used in Study
```

H0R5E:
@ssolson & @ryancoe, I just did a force push on this to rebase it onto the current trunk. You should delete you local copies of this branch and re-check it out, to avoid any nasty conflicts.

ryancoe:
I gambled and just pull/merged... worked fine for me

H0R5E:
@ryancoe & @ssolson, I think this is ready for review. There are 3 issues to resolve / check. These are:

1. For the resampling I have chosen to use the maximum error in the spectral density, normalised by the maximum density in the spectrum, as the metric. This performs well (as seen in the graphs I posted to #29) and is a nice metric because you know the maximum that each original sample will be out by. 
2. I didn't conform to using integer values of `dw` for the resampling as it can behave oddly if you are resampling a spectrum that also doesn't use integer values (although `dw` is still regular). From my testing, I don't think this is an issue, but it would be good to confirm it.
3. The matlabdomain Sphinx plugin can't cope with the `arguments` keyword at the moment. I put a [PR](https://github.com/sphinx-contrib/matlabdomain/pull/111) on their repo, but until that is incorporated, or we create a personal version of the package, we can't publish the docs for this PR.

Regarding the level of review, I would, again, suggest that perhaps addressing the issues above is enough given how long this PR has been open, and checking that the changes to the example make sense. Let's discuss.

ssolson:
@H0R5E Thanks for all your work on the PR. I have been looking over some of your changes this morning and per usual you have done a great job. I will start my review of this today.

H0R5E:
Thanks Sterling. Here are the test results:

[test_results.pdf](https://github.com/SNL-WaterPower/WecOptTool/files/4879594/test_results.pdf)

ssolson:
@H0R5E I messed up by merging our branches (I didn't want to keep my changes) then further solidified this by misusing revert. Can you make a PR of your current version into my subSampleW branch? I am having trouble identifying what I need to remove/ keep with the large change to the structure.

ssolson:
> @H0R5E I messed up by merging our branches (I didn't want to keep my changes) then further solidified this by misusing revert. Can you make a PR of your current version into my subSampleW branch? I am having trouble identifying what I need to remove/ keep with the large change to the structure.

I did a "manual reset" by deleting the repository and copying in the files from your last commit. This is probably wrong but allows me to currently continue my review.

H0R5E:
Yeah, it's not easy to deal with a rebased shared remote branch. Overwriting the files is a perfectly valid way to recover and all sins will be erased when it's merged into the trunk anyway.

For reference, if you need to do something similar (and assuming you don't have local changes), you can follow this sequence:

```
> git fetch
> git checkout <branchname>
> git reset --hard origin/<branchname>
```

H0R5E:
Frequencies must now be integer multiples of the step.

ssolson:
> Frequencies must now be integer multiples of the step.

f52bd7b breaks all the sea state tests and 1 RM3 test.

H0R5E:
> [f52bd7b](https://github.com/SNL-WaterPower/WecOptTool/commit/f52bd7b2b9f0b71b5aa61f8def4aa697154a4fa5) breaks all the sea state tests and 1 RM3 test.

Working on it.

ssolson:
When I resample by stepsize why does the title say "Max" Resampling error?

image

ssolson:
Mat I think the way you have organized the sea states into a class is brilliant. I love that I can load, verify, modify, and plot sea states all in one place. The documentation is also brilliant, easy to read, and follow. I feel like anything I may want to change is knit picky and I approve this PR in favor of moving on to more important development tasks. I have found a couple of typos and I will Americanize a couple of words but other than that this is ready to go.

H0R5E:
Thanks Sterling. I've just changed the travis documentation build instructions and the sphinx environment setup instructions in the README so that the docs will compile when this is merged and so others can compile it too. Are you going to make those changes you suggested now, or in another PR?

H0R5E:
> When I resample by stepsize why does the title say "Max" Resampling error?

Because it's using the same metric to calculate the error as the resampleByError method, which is the maximum absolute error in spectral density, normalised by the max (from the original samples).

ryancoe:
@H0R5E - This looks great and I generally agree that we should merge this and deal with small stuff later.

The two more major things:

1. I would prefer to change (maybe you have a good reason not to that I haven't thought of) is that the fields of the `SeaState` class (e.g., `SS.S`) are read-only. This is a bit annoying for me, as I sometimes like to, e.g., set all of the spectral density values to zero except one for debugging.

2. I had trouble calling some of the methods. For example, I wanted to see what `getSpecificEnergy` did, so I tried:

```
>> SS.getSpecificEnergy()
Error using WecOptTool.types.SeaState.getSpecificEnergy
Invalid argument list. Not enough input arguments. Function requires 1 positional input(s).
```

Looking at the unit tests, I figured out that I must use the following, which I thought was unintuitive. Is there something I'm missing here?

```
SS.getSpecificEnergy(SS)
```

---

Other minor things to maybe note for later work:

 - might prefer to have the `plot` method use only one figure with subplots (could be an option)
 - combine methods `resampleByStep` and `resampleByError` into one method controlled via arguments
 - would love to have a page on the documentation site that talks about this in more detail
 - (general comment) it would be great if we could not replicate documentation for sphinx and MATLAB... not sure of a solution yet (https://github.com/sphinx-contrib/matlabdomain/issues/105#issuecomment-654983322)

H0R5E:
> 1. I would prefer to change (maybe you have a good reason not to that I haven't thought of) is that the fields of the `SeaState` class (e.g., `SS.S`) are read-only. This is a bit annoying for me, as I sometimes like to, e.g., set all of the spectral density values to zero except one for debugging.

So this was a specific design decision I made about Data classes being invariant snapshots from the point of their creation (as per #126). So the idea is that there is a guarantee that the data in Data subclasses can not be modified accidentally and therefore you can be confident as a user that it's giving you what you expect. Basically this means you have to be deliberate about modifying them by converting to struct and then recreating if you want, i.e.

```
>> S = WecOptLib.tests.data.exampleSpectrum();
>> SS = WecOptTool.types("SeaState", S);
>> Snew = struct(S);
>> S.S = 0;
>> SSnew = WecOptTool.types("SeaState", Snew);
```

In summary, the main idea behind this is to reduce the risk of accidental modification to data that may be used later in the objective function.

H0R5E:
> 2\. I had trouble calling some of the methods. For example, I wanted to see what `getSpecificEnergy` did, so I tried:
> 
> ```
> >> SS.getSpecificEnergy()
> Error using WecOptTool.types.SeaState.getSpecificEnergy
> Invalid argument list. Not enough input arguments. Function requires 1 positional input(s).
> ```
> 
> Looking at the unit tests, I figured out that I must use the following, which I thought was unintuitive. Is there something I'm missing here?

These are static functions which work on the raw structs. It says this in the help:

```
>> help WecOptTool.types.SeaState.getSpecificEnergy
  Calculates the specific energy of the given spectra struct 
  array
 
  Arguments:
     S (struct):
         struct array that satisfies the
         :mat:meth:`+WecOptTool.+types.SeaState.checkSpectrum` 
         method
     options: name-value pair options. See below.
 
  The following options are supported:
 
     g (optional, float):
         Acceleration due to gravity, default = 9.81 
         m/s\ :sup:`2`.
     rho (optional, float):
         Water density, default 1028 kg/m\ :sup:`3`.
 
  Returns:
      array: specify energy per spectra [J / m\ :sup:`2`]
```

We could break these off into a separate module maybe, rather than include them in the class?

H0R5E:
> * combine methods `resampleByStep` and `resampleByError` into one method controlled via arguments

Not sure I see the value is this, because your going to have a more complex syntax for the main call, something like:

```
>>> S = WecOptLib.tests.data.example8Spectra();
>>> SS = WecOptTool.types("SeaState", S,           ...
                          "resample",                               ...
                          "resamplstepLength", 0.1,       ...
                          "trimFrequencies", 0.01, ...
                          "extendFrequencies", 4);
```

So, I don't think this is any easier to understand, unless you meant something else? It's probably mute with the resampling by energy anyway as every step length will have zero error, so you will only resample by step size in the future.

ryancoe:
- before merge
 	- add examples to each static function in SeaState
 	- add specific energy property to object
 - after merge
 	- consider moving static methods to regular methods (and creating new SeaStates on call)
 	- will add documentation on SeaState class (can add to this later for empirical sea states)
 	- add more properties that get populated on instantiation (moments, periods, etc.)
 	- consider making plots in a single figure

PR #158, opened 2020-07-14
Documentation: Update WecOptTool flowchart 

### LINKED ISSUE #96: WecOptTool Flow Chart
I took an hour to recreate the important aspects of the WecOptTool flow chart in LaTex Tikz. I didn't finish but I wanted to see if Ryan hated it enough to tell me to stop anyway. 

Edit link is here: https://www.overleaf.com/8641328196bhrsbhbfmbrr

ssolson:
## Description

This PR updates the documentation to include the latest WecOptTool flowchart

Fixes #96

## Checklist:
**Only documentation was changed**

~~- [ ] Added the results of running the test suite (in pdf form)~~
    
~~- [ ] All new files contain the GPL header~~
~~- [ ] If `example.m` has been modified, the content / line numbers in
  `docs\user\example.rst` are still valid or have been fixed~~


ryancoe:
Remove blank space below figure

If possible, match figure font to Sphinx

H0R5E:
I can't say definitely until I get home and test a build, but I suspect it's fine.

What's not ok for me is one sentence per line in the rst file. If you look at a rst linter like doc8 https://pypi.org/project/doc8/ lines should have a max length (79) to aid readability in editors.

ryancoe:
> What's not ok for me is one sentence per line in the rst file. If you look at a rst linter like doc8 https://pypi.org/project/doc8/ lines should have a max length (79) to aid readability in editors.
> 
> EDIT: Another benefit is that it's easier to find errors when only a limited number of characters are allowed on a line. Same goes for latex.

This is definitely a bit of a judgement call, and I probably shouldn't have made the changes so broadly without discussing first. I always use one line per sentence in LaTex; I don't write much `rst`. My editor/linter (Sublime Text) actually works better without the excess line breaks. What I don't enjoy doing is re-breaking the lines when making edits to get everything to <79 chars... I see the `doc8` point, but some of those are subjective (e.g., tabs vs. spaces; no newline at end of file). Let's discuss tomorrow.

H0R5E:
Won't be on the call but:

https://github.com/randy3k/AutoWrap

I must say that for me it makes the source files pretty unreadable as I have soft word wrapping turned off in my editors.

H0R5E:
A discussion on this about latex:

https://tex.stackexchange.com/questions/325505/best-practices-for-source-file-line-lengths

ryancoe:
@H0R5E - Why do you always have to make sure compelling and well supported arguments ;)

H0R5E:
I can find counter examples to using hard wraps such as https://docs.pylonsproject.org/projects/docs-style-guide/ and there are good arguments for soft wrapping like it's easier to do edits on GitHub. 

I have to say that i am deeply biased on this one, the so it's difficult to be objective, sorry. It's not the end of the world if you want to do a sentence per line, TBH. I'll just have to learn how to turn on soft wrapping in NP++.

H0R5E:
Coming back to this. I think my biggest gripe with using soft wrapping for readability purposes (which is the principle reason, I would say, for using hard wrapping) is that when we need to use more width for something like tables, it looks totally awful when they are soft wrapped. For me, hard wrapping is the only way to format a plain text file so that it's readable in all circumstances.

H0R5E:
Yeah, it compiles OK, but I found a typo caused from by an incorrect break inside an object link. I haven't checked all of it for similar issues.

PR #166, opened 2020-09-03
Fix No Warning Assertion on Plot Tests

H0R5E:
## Description

If hardware support is not available on the testing platform then some plots may throw a warning so the verifywarningfree assertion has been removed from plot tests.

## Checklist:

[test_results.pdf](https://github.com/SNL-WaterPower/WecOptTool/files/5167966/test_results.pdf)


ssolson (commented):
[test_results.pdf](https://github.com/SNL-WaterPower/WecOptTool/files/5176728/test_results.pdf)

PR #181, opened 2020-09-19
Fix details not being unrolled when clicking on summary in Chrome

H0R5E:
## Description

Chrome does not treat \<a\> tags inside a \<summary\> tag the same way as Firefox, such that the details tag will not unroll when the summary is clicked. This commit swaps the \<a\> tags for a custom tag \<clickme\> that colours the summaries like links, without being links. Chrome now acts as expected.

Found a couple of small fixes to the API docs which I have added to this PR.


H0R5E:
OK, I'm going to merge this as I don't think it's going to break anything.

PR #183, opened 2020-09-24
WaveBot: example script readability and comments

ryancoe:
## Description
Minor changes to improve the readability of the WaveBot example scripts and to add comments where appropriate to make the higher level meanings of steps within the code more clear.

## Checklist:

### DIFF
--- a/examples/WaveBot/WaveBot_caseA.m
+++ b/examples/WaveBot/WaveBot_caseA.m
@@ -28,8 +28,6 @@
 %     You should have received a copy of the GNU General Public License
 %     along with WecOptTool.  If not, see <https://www.gnu.org/licenses/>.
 
-clear controlParams r
-
 %% define sea state of interest
 
 dw = 0.3142;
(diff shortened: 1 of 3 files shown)


ryancoe:
All tests pass: [test_results.pdf](https://github.com/SNL-WaterPower/WecOptTool/files/5285490/test_results.pdf)

ryancoe:
@H0R5E - If you find this ready for merge, please go ahead and execute.

H0R5E:
Still good without the PCT.

[test_results_no_pct.pdf](https://github.com/SNL-WaterPower/WecOptTool/files/5293262/test_results_no_pct.pdf)

H0R5E (approved):
So, the only thing I would say is that "Monte-carlo" is a bit of a stretch to describe the simulation of the pre-selected radii in case B, given there is no randomness to the selection of the values.

ryancoe:
Monte-Carlo: yes, good point... dealing with a deterministic system in this case it doesn't seem that random sampling would add anything... probably best to leave it at "brute force" though

PR #189, opened 2020-09-29
WaveBot case A: x-axis integer multiples of fund. freq.

ryancoe:
## Description
It is nice to label the x-axis for this case with integer multiples of the fundamental frequency to better highlight the harmonics. I've also remove the `xticklabel`s from the first row of axes.

94573158-0bd29280-022f-11eb-9665-40ed1c0cc196

## Checklist:

### DIFF
--- a/examples/WaveBot/Performance.m
+++ b/examples/WaveBot/Performance.m
@@ -16,15 +16,18 @@ classdef Performance < handle
     
     methods
         
-        function plotTime(obj ,t)
+        function plotTime(obj, t, options)
             
-            if nargin < 2
-                trep = obj(1).getRepeatPer();
-                t = 0:0.05:trep;
+            arguments
+                obj
+                t = 0:0.05:obj(1).getRepeatPer()
+                options.Interpreter = 'none'
+                options.FontSize = 11
             end
             
             fig = figure('Name','Performance.plotTime');
-            fig.Position = fig.Position.*[1 1 1 1.5];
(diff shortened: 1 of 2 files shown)


ryancoe:
All tests pass: [test_results.pdf](https://github.com/SNL-WaterPower/WecOptTool/files/5299871/test_results.pdf)

H0R5E:
@ryancoe, I can't run this case, so you should probably ask Sterling to take a look at it.

ryancoe:
@H0R5E - What's preventing you from running this case?

ssolson:
Ryan Case C uses the "Global optimization toolbox" not the optimization toolbox. So I can look into this.

ryancoe:
This is `caseA` -- my mistake in labeling the PR.

H0R5E:
So, I have pushed some of the functionality in this case to the Performance class plot functions (such as setting the interpreter and font size) and tidied up a bit. This also sets the font of the x and y axis labels to latex, which I think is how it should be. 

The other change I made here was to change 1 \omega to simply \omega in the x-axis tick labels. It's unusual to put 1 * x rather than just x.

If you feel OK with these changes, then I'd be happy for this to be merged now.

ryancoe:
All tests pass and looks good to me: [test_results.pdf](https://github.com/SNL-WaterPower/WecOptTool/files/5385595/test_results.pdf)

PR #193, opened 2020-10-20
Shape Study

ssolson:
## Description

Performs a geometric shape study characterizing relationships between a device shape and its response using the WecOptTool.

* Compares QoIs for constant volume and natural frequency for varying aspect ratios in a Jonswap spectrum.

image

In progress:
Investigating Power and Constant force excitation

## Checklist:

- [ ] All new files contain the GPL header
- [ ] If `examples/RM3/optimization.m` has been modified, the content / line 
  numbers in `docs/user/optimization.rst` are still valid or have been fixed


ssolson:
@ryancoe I have put together a diagram of what I am doing to better understand what I need to be doing. There are a couple of points from our last discussion that I could use clarification on. I'll reach out about a potential call to discuss.

image

ryancoe:
Per our conversation today, it would probably be beneficial to back-calculate the wave elevation (eta) required to make the constant power and constant force excitations.
wecModelingAndControlWorkflow

PR #194, opened 2020-10-22
update publications, references, etc.

ryancoe:
Update the DOI for the TechRxiv preprint and note acceptance of paper by _Journal of Ocean Engineering and Marine Energy._

closes #190 #172 

EDIT: additionally now updated [References](https://snl-waterpower.github.io/WecOptTool/user/references.html) page to include project publications

PR #197, opened 2020-11-10
Change Citing WecOptTool reference to published paper

H0R5E:
## Description

#196 for the Citing WecOptTool page.

PR #198, opened 2020-11-10
fix #194

ryancoe:
We forgot to update [`4cite.rst`](https://github.com/SNL-WaterPower/WecOptTool/blob/master/docs/user/4cite.rst)


ryancoe:
@H0R5E got to it first with #197

Issue #153, opened 2020-07-09
Document SeaState class example


ryancoe:
Programmatic steps to create, alter, plot, etc. `SeaState` objects.

ssolson:
From our discussion, this morning documentation of how the sea state is applied to BEM vs Control solver was brought up. I offered to start a flow chart for this and I believe it fits into this blanket SeaState class if not I lmk and I will create a separate issue.

My Idea was to add detail to the current flowchart explaining. I have modified the original flowchart changing the location of some of the blocks to better demonstrate this process. This has the advantage of potentially allowing a common reference point with more details as needed but if I need to move the blocks based on the process I want to explain better this could be a slippery slope.

As you can see I have greyed out the areas not being considered. I moved the sea states above the controller and added the "Representative Spectra". Assume this is a draft and I will clean up the lines etc and that this will have its own page in the documentation explaining the detail.

![image](https://user-images.githubusercontent.com/13438942/88333152-f4d26a00-ccec-11ea-8af9-cd9ab5acc5a0.png)


My question to you guys before I go further would be should I make a separate flowchart for this process or is this template good?

ryancoe:
@ssolson will come back to this later, general idea: don't make different versions of the diagram, try to use code/plots to illustrate the distinct frequency vectors

H0R5E:
Suggest that we consider constructing this documentation as part of an "Error Handling" page and include information about #203 (assuming it's merged) alongside.

PR #203, opened 2020-11-24
Calculate standard error on PS results

### LINKED ISSUE #56: Ensure sufficient number of phase realizations in PS
> We use multiple realizations to deal with phase-dependent nature of a nonlinear model. Since the pseudo spectral results are nonlinear, they are dependent on the phasing of the waves. Thus, we average over 5 realizations - this was previously found to be sufficient, but I think it warrants another look just to make sure. I believe we also considered using a while loop and looking for convergence...

_Originally posted by @ryancoe in https://github.com/_render_node/MDIzOlB1bGxSZXF1ZXN0UmV2aWV3VGhyZWFkMjQ1MjI5MTYzOnYy/pull_request_review_threads/more_comments_

H0R5E:
## Description

This commit uses the central value theorem to facilitate calculation of the error on the mean results of the PS controller, using the standard error from a normal distribution.

Two new functions are added to the WecOptTool.math package called standardErrorMeasure and standardErrorReduce, the first of which measures the standard error after a given number of samples and the second of which keeps sampling until a desired level of error is found.

An example script to show how the standard error varies with the stroke of the RM3 device using the PS controller is added to the RM3 example directory called PSError.m. The results of this simulation are shown here:

RM3_PS_standard_error

Fixes #56 

## Checklist:

- <s>[ ] If `examples/RM3/optimization.m` has been modified, the content / line 
  numbers in `docs/user/optimization.rst` are still valid or have been fixed</s>
- [x] Add docstrings
- [x] Add tests
- <s>[ ] Add something to the documentation?</s> -> Do as part of #153


H0R5E:
Hi Ryan,

I implemented your algorithm, but I'm not sure it's behaving the way you expected it to do. The diagram below shows a simulation with the tolerance set to `1e-6`.

ryan_algorithm

I don't think this is converging as you were expecting, but I'm not really surprised because of the random numbers.

I'm also not sure what the tolerance is telling you. I set the tolerance to `1e-4` and ran the algorithm 10 times. The range on the results was `69039.8` Watts, but `1e-4 * max(power)` is `395.2` Watts, so the tolerance is kind of misleading.

ryan_algorithm_2

How do you want to proceed?

H0R5E:
I've added a metric which calculates the sum of the standard errors over the sum of the mean. It produces results as such:

RM3_PS_standard_error_of_mean

I'll do the reverse and set a limit on the error and show how many iterations are required.

H0R5E:
OK, so here is the same situation recording the number of iterations required:

RM3_PS_standard_error_N

So, I really think the question is whether the use wants to set a particular accuracy and wait for the realisations, or choose a number of realisations and suffer the inaccuracy. I suggest we offer both options.

H0R5E:
Just a note that this would need to be updated if #170 is merged first and visa-versa.

H0R5E:
As part of the testing process I noticed that when summing the errors the results were somewhat better than expected. Turns out, that's because summing the standard error is not the correct approach and taking the 2norm is the correct way to go. I've implemented that now and the results are more consistent now.

H0R5E:
Add option when reducing error for max samples - return error?

H0R5E:
@ryancoe

PR #207, opened 2021-01-12
Flapping device

### LINKED ISSUE #156: Flapping device

H0R5E:
## Description

Continuation of #170

Fixes #156

## Checklist:

<s>- [ ] If `examples/RM3/optimization.m` has been modified, the content / line 
  numbers in `docs/user/optimization.rst` are still valid or have been fixed</s>


ryancoe:
Can you put together a MWE along the lines of this?

`[Aw, Bw, Fexc, P_cc] = flapper_mwe(depth)`

H0R5E:
So, these are the results from modifying the depth of Giorgio's flap from 15m (minimum depth) to 30m in 5m intervals.

```
                AvgPow        |MaxPow|     PowTHD_dBc    MaxTheta    MaxOmega    MaxTauPTO 
              ___________    __________    __________    ________    ________    __________

    15.0 m    -2.7235e+05    1.2862e+06       NaN        0.051194    0.028945    4.8244e+07
    20.0 m    -3.0834e+05     3.752e+06       NaN         0.18056     0.10012    6.1384e+07
    25.0 m     -3.378e+05    6.8582e+06       NaN         0.26963     0.14805      8.24e+07
    30.0 m    -3.5975e+05    1.1169e+07       NaN         0.36094     0.19639    1.0558e+08
```

model_results

The input values for added mass, damping and excitation are shown in the plots below:

added_mass
damping
excitation

H0R5E:
I never got a chance to discuss it in our meeting, but I've been looking at this some more, and my current thoughts are that this is either a quirk of the maths (and therefore is maybe a matter of adding a constraint?) or something fishy in the outputs of NEMOH. 

So, fundamentally, for the CC controller (and unconstrained PS, by design) the optimal complex velocity is the ratio of the excitation force to the real part of the impedance matrix (i.e . the radiation damping). So for the result, the absolute size of these values is irrelevant. What is happening in the NEMOH results is that as the depth of the flap is increasing, the radiation damping is getting smaller faster than the excitation force, and thus the optimal velocity and power are increasing.

It seems to me that this is going to quickly exceed the upper limit on the extractable power from the wave, so I'm not sure what assumption is being broken to allow this happen. Is there some relationship between the radiation damping and the excitation force that is being broken by NEMOH? It would be interesting to see what capytaine does with this case, also.

